PR #1, opened 2019-06-25
Make use of wagtail-headless-preview

zerolab:
Switch to using [Wagtail Headless Preview](https://github.com/torchbox/wagtail-headless-preview) and tweaked the README


zerolab:
Change of plan. As discussed, will move the live headless bits into the headless preview package

PR #12, opened 2019-07-29
Feature/circle ci

jafacakes2011:


jafacakes2011:
The issue is with the database. We need to either remove the DB and build a new one for the test. Or work out how to make the DB work with the test script

jafacakes2011:
We are still getting the DB issue, but LFS is now available in the test scripts so we should be closer.

jafacakes2011:
FIXED!

zerolab:
Nice one! I suggest merging this with squashed commits so the history is clean and doesn't contain all the intermediary attempts

PR #28, opened 2019-10-25
Feature/fields refactor

NathHorrigan:


zerolab:
LGTM. 👍for the embed test. Noticed there's no test for `GraphQLPage` which is probably a good idea to add here

PR #30, opened 2019-11-01
Reorganize file structure so can use manage.py and test easier

NathHorrigan:
All files have been moved inside of an example project which means I can make use of the Django projects `manage.py` for creating migrations. 

Setup.py has been reconfigured so output is the same.


NathHorrigan:
File structure now like so:

Screenshot 2019-11-01 at 12 43 42

zerolab:
I am not convinced by this change.
A better way would be to follow what Wagtail does with its testapp -- https://github.com/wagtail/wagtail/tree/master/wagtail/tests/testapp

The way I see it, the `example` app is very much for tests so we should aim to keep it as that. I understand you also use it locally for demos and whatnot, but that is outside its remit.

Happy to be convinced otherwise :)

PR #36, opened 2019-11-27
Fix 2.7 requirements

### LINKED ISSUE #35: Support for Wagtail 2.7
I know Wagtail 2.7 is quite fresh still, but just wondering whether it would be possible to add support? Pipenv is complaining about incompatible versions in dependencies when trying to install `wagtail-grapple` into a project with Wagtail 2.7.

zerolab:
Should fix #35

requirements.txt had Wagtail <2.8, but setup.cfg and the README still references 2.7

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ Getting Grapple installed is designed to be as simple as possible!
 ### Prerequisites
 ```
 Django  >= 2.2, <2.3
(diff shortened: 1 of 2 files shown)


zerolab:
Depends on https://github.com/torchbox/wagtailmedia/pull/63

PR #44, opened 2020-01-10
Allow turning off graphene’s ‘auto_camelcase’

### LINKED ISSUE #40: request: setting to disable auto-conversion from snake_case to camelCase
Currently, the graphene schema is initialized as follows:

https://github.com/torchbox/wagtail-grapple/blob/c35408ad3e6a5e273d1e2316c0f3c988f5064501/grapple/schema.py#L47-L49

This automatically converts model attribute names from e.g. `url_path` to `urlPath`. While some people might like this because it conforms to the convention of camel-casing names in javascript, it also fails the [grep test](http://jamie-wong.com/2013/07/12/grep-test/) which makes things harder for someone exploring a new codebase and learning a new stack.

Recommend adding a new setting, e.g. `GRAPPLE_AUTO_CAMELCASE` which can be set to `False` and [passed to graphene](https://docs.graphene-python.org/en/latest/types/schema/) through the `auto_camelcase` parameter.

NathHorrigan:
Fixes #40

Stops Graphene automatically converting Object fields from underscore case to camel case so:

```
{
  pages {
    ...on BlogPage {
      title
      bookFile {
        createdAt
      }
    }
  }
}
```

becomes...

```
{
  pages {
    ...on BlogPage {
      title
      book_file {
        created_at
      }
    }
  }
}
```

You simply just add `GRAPPLE_AUTO_CAMELCASE=False` to `base.py` settings file.

### DIFF
--- a/docs/getting-started/examples.rst
+++ b/docs/getting-started/examples.rst
@@ -70,6 +70,6 @@ something like:
 
 **Next Steps**
 
-  * :doc:`../general-usage/types`
+  * :doc:`../general-usage/graphql-types`
   * :doc:`../general-usage/preview`
 
--- a/docs/getting-started/installation.rst
+++ b/docs/getting-started/installation.rst
@@ -43,12 +43,16 @@ Add the GraphQL urls to your ``urls.py``:
    ]
 
 Done! Now you can proceed onto configuring your models to generate
-GraphQL types that adopt their stucture.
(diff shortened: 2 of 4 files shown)


ruisaraiva19:
> Needs documenting the change. and ideally a small test

Documentation and test added on #58

PR #47, opened 2020-03-03
Fix ListBlock streamfield type.

### LINKED ISSUE #33: Ability to add alongside existing graphql implementation?
Hey, 

Love the potential of this project, and we'd like to explore using it. However, we'd need to migrate away from our current graphene implementation, and this could take some time. Is there a way we could start including grapple as an alternate endpoint? 

I tried giving it its own namespace in urls.py, but setting the schema as instructed in settings.py also sets the schema for our existing graphql endpoint, which doesn't let them live side by side? Any advice? 

If we get this working, I'd be happy to do a PR to the docs as an example, as I'm sure this could be a common enough situation and could help make it easier for others to adopt this library!

### LINKED ISSUE #38: ListBlock implementation incomplete?
I'm trying to create a simple Gallery Block working for use within a StreamField field. For example:

```python
body = StreamField([
    ('text', MarkdownBlock()),
    ('text_image', TextImageBlock()),
    ('image', ImageChooserBlock()),
    ('url', URLBlock()),
    ('video', VideoBlock()),
    ('code', CodeBlock()),
    ('gallery', ListBlock( ImageChooserBlock() )),
])
```

> The item at the bottom.

Within the GraphQL response when querying, `items` is being returned as an object rather than array of items and the object returned is actually itself.

Is this implementation complete? I'm by far a Python expert but could it be to do with the below line of code?

https://github.com/torchbox/wagtail-grapple/blob/cb17c951b7b47a014aebccdf15358ac241acd86c/grapple/types/streamfield.py#L329

The items resolve to itself?

NathHorrigan:
Finishes implementation of ListBlock Streamfield type.
- Fixes #38 

Streamfield Model Definition:
```python
body = StreamField([
    ('pages', blocks.ListBlock(blocks.PageChooserBlock()))
])
```

Query:
```graphql
{
  pages {
    ...on BlogPage {
	body {
        ...on ListBlock {
          field
          id
          items {
           ...on PageChooserBlock {
              page {
                id
                title
              }
            }
          }
        }
      }
    }
  }
}
```

Query Response:
```json
{
  "data": {
    "pages": [
      {},
      {},
      {
        "body": [
          {
            "field": "pages",
            "items": [
              {
                "page": {
                  "id": "3",
                  "title": "Home"
                }
              },
              {
                "page": {
                  "id": "4",
                  "title": "Blog"
                }
              }
            ]
          }
        ]
      },
      {
        "body": []
      }
    ]
  }
}
```


zerolab:
The changes LGTM. Not sure how it solves #33 though (would be good to update docs, if this is the case). Needs a test that confirms the output (as outlined in the PR description)

NathHorrigan:
Sorry my bad, I meant #38.

zerolab:
👍

PR #51, opened 2020-03-12
Add non-nullable support to fields

NathHorrigan:
This feature was requested and add's support for https://docs.graphene-python.org/en/latest/types/list-and-nonnull/#nonnull

This is used like so:
```
GraphQLString("date", required=True),
```

or for collections:
```
GraphQLCollection(
            GraphQLForeignKey, "related_links", "home.blogpagerelatedlink", required=True, item_required=True
),
```

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -31,12 +31,11 @@ include_package_data = true
 packages = find:
 install_requires =
     Django>=2.2, <2.3
-    wagtail>=2.5, <2.8
+    wagtail>=2.5, <2.9
     graphene-django==2.7.1
     graphql-core==2.2.1
-    colorthief
     channels==1.1.8
     asgi_redis
     graphql_ws
-    scour==0.37
     wagtail-headless-preview
+    graphene-django-optimizer==0.6.1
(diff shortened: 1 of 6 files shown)


setup.cfg
```diff
@@ -31,12 +31,11 @@ include_package_data = true
...
+    wagtail>=2.5, <2.9
     graphene-django==2.7.1
     graphql-core==2.2.1
-    colorthief
     channels==1.1.8
     asgi_redis
     graphql_ws
-    scour==0.37
     wagtail-headless-preview
+    graphene-django-optimizer==0.6.1
```
ruisaraiva19: Is this packages being used anywhere?

setup.cfg
```diff
@@ -31,12 +31,11 @@ include_package_data = true
 packages = find:
 install_requires =
     Django>=2.2, <2.3
-    wagtail>=2.5, <2.8
+    wagtail>=2.5, <2.9
```
ruisaraiva19: Nice, Wagtail 2.8.x compatibility!

setup.cfg
NathHorrigan: No, That found it's way in thank you!

PR #54, opened 2020-03-25
(Fix #48) ImageChooserBlock resolving to an integer within StreamBlocks

### LINKED ISSUE #48: StreamBlock implementation issue
Hi there,

I am receiving the following exception when querying an image inside of a StreamBlock (see use case and query below):

> Exception: Received incompatible instance "1".

The issue appears to be due to the fact that the StreamBlock implementation in streamfield.py uses the value from StreamChild.value.stream_data to resolve the stream blocks.  For ImageChooserBlock, it looks like the value in stream_data resolves to its ID, and not the Image object, as in the following (at least this is so for the Wagtail core Image):

> {'type': 'photo', 'value': 1, 'id': 'd60d57b2-5bb5-4eec-9f55-1a70b4fb5ad8'}

This dict is an example of what you get for the “field” variable in the for loop on line 153:

```
    def resolve_blocks(self, info, **kwargs):
        stream_blocks = []
        for field in self.value.stream_data:
            block = self.value.stream_block.child_blocks[field["type"]]
            if not issubclass(type(block), blocks.StructBlock):
                value = block.to_python(field["value"])

            stream_blocks.append(StructBlockItem(field["type"], block, field["value"]))
        return stream_blocks
```

This isn’t a problem for “scalar” blocks, because the value in the stream_data dict generally works.

From the digging I’ve done so far, in order to properly resolve the stream blocks for an object like Image, the values need to resolve to the image objects, which can be accessed as items in StreamChild.value, like in your ListBlock implementation for #47 .  So something like this could work, though I could be missing other pieces of the implementation:

```
    def resolve_blocks(self, info, **kwargs):
        stream_blocks = [StructBlockItem(item.id, item.block, item.value) for item in self.value]
        return stream_blocks
```

I didn't look too deeply into the StructBlock implementation, so I don't know if there's a potential issue there as well.

The following basic use case is what got me into trouble in the first place:

```
class ArticlePage(Page):
    body = StreamField([
        ('section', blocks.StreamBlock(
            [
                ('text', blocks.RichTextBlock()),
                ('photo', ImageChooserBlock()),
            ]
        )),
    ])

    content_panels = Page.content_panels + [
        StreamFieldPanel('body'),
    ]

    graphql_fields = [
        GraphQLStreamfield('body'),
    ]
```

With a query like:

```
{
  pages {
    ... on ArticlePage {
      body {
        ... on StreamBlock {
          blocks {
            field
            blockType
            ... on RichTextBlock {
              value
            }
            ... on ImageChooserBlock {
              image {
                title
                src
              }
            }
          }
        }
      }
    }
  }
}
```

Which gives a result like:

```
{
  "errors": [
    {
      "message": "Received incompatible instance \"1\"."
    }
  ],
  "data": {
    "pages": [
      {
        "body": [
          {
            "blocks": [
              {
                "field": "photo",
                "blockType": "ImageChooserBlock",
                "image": null
              },
              {
                "field": "text",
                "blockType": "RichTextBlock",
                "value": "<p>Lectus mauris ultrices eros in cursus turpis massa tincidunt. Cursus euismod quis viverra nibh cras pulvinar mattis. Elit eget gravida cum sociis natoque penatibus et magnis dis. Nunc aliquet bibendum enim facilisis gravida neque convallis a. Nisi est sit amet facilisis magna etiam tempor orci. Aliquet eget sit amet tellus cras adipiscing enim eu. Diam quam nulla porttitor massa id neque aliquam vestibulum. Sit amet purus gravida quis blandit turpis cursus in hac. In hac habitasse platea dictumst vestibulum rhoncus est. Volutpat sed cras ornare arcu dui.</p>"
              }
            ]
          }
        ]
      }
    ]
  }
}
```

janbaykara:
This PR pipes the correct value for `ImageChooserBlock` through to the `StreamBlock.resolve_blocks` resolver, so that the consumer can access image path information etc.

This fixes #48 which I was also experiencing, where such blocks in StreamBlock would resolve to an integer rather than to the image object.

Screenshot 2020-03-25 at 18 07 09

### DIFF
--- a/grapple/types/streamfield.py
+++ b/grapple/types/streamfield.py
@@ -155,10 +155,11 @@ class StreamBlock(StructBlock):
         for field in self.value.stream_data:
             block = self.value.stream_block.child_blocks[field["type"]]
             value = field['value']
-            if issubclass(type(block), block.ChooserBlock):
+            if (
+              issubclass(type(block), wagtail.core.blocks.ChooserBlock)
+              or not issubclass(type(block), blocks.StructBlock)
+            ):
                 value = block.to_python(value)
-            elif not issubclass(type(block), blocks.StructBlock):
-                value = block.to_python(field["value"])
 
             stream_blocks.append(StructBlockItem(field["type"], block, value))
         return stream_blocks
--- a/grapple/types/streamfield.py
+++ b/grapple/types/streamfield.py
@@ -137,7 +137,10 @@ class StructBlock(graphene.ObjectType):
         stream_blocks = []
         for name, value in self.value.items():
             block = self.block.child_blocks[name]
-            if issubclass(type(block), blocks.ChooserBlock):
+            if (
+              issubclass(type(block), wagtail.core.blocks.ChooserBlock)
+              and hasattr(value, 'id')
+            ):
                 value = block.to_python(value.id)
             elif not issubclass(type(block), blocks.StreamBlock):
                 value = block.to_python(value)
--- a/grapple/types/streamfield.py
+++ b/grapple/types/streamfield.py
@@ -137,7 +137,9 @@ class StructBlock(graphene.ObjectType):
         stream_blocks = []
         for name, value in self.value.items():
             block = self.block.child_blocks[name]
-            if not issubclass(type(block), blocks.StreamBlock):
+            if issubclass(type(block), blocks.ChooserBlock):
+                value = block.to_python(value.id)
+            elif not issubclass(type(block), blocks.StreamBlock):
                 value = block.to_python(value)
 
             stream_blocks.append(StructBlockItem(name, block, value))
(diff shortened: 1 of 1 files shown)


zerolab:
Hey @janbaykara,

Thank you for this! Can you please add a quick test so any future changes do not break your fix?

janbaykara:
I've widened this fix upon noting a similar error with PageBlockChoosers in StructBlocks, upon realising the error most likely relates to the `to_python` method and its use of primary key lookup:

https://github.com/wagtail/wagtail/blob/4314f3d1a1419f7a184c858f8cc8afa1dbfb3d5c/wagtail/core/blocks/field_block.py#L634

> Hey @janbaykara,
> 
> Thank you for this! Can you please add a quick test so any future changes do not break your fix?

Shall do shortly.

janbaykara:
(Definitely may have overstretched with that last commit; leaving this thread open to discuss possible solutions and make notes on tests.)

NathHorrigan:
This is great, Thanks @janbaykara.
Once we have tests I'll merge this :)

janbaykara:
Apologies, I think I may have bitten off more than I can chew here. Not entirely clear to my sleepy weekend brain how to wrangle the factory system into setting up the block structure I need to test things. Will have to read into this with some time and focus at a future date.

janbaykara:
I mentioned in #55 — I'm uncertain when I'll have time to learn what testing looks like for this, so if someone wants to dip in on that, please do.

tbrlpld:
Just ran into both these issues today. Not sure of how much help I can be with the tests, as I am not familiar with the implementation at all so far.

NathHorrigan:
Closed in favour of #82

PR #56, opened 2020-03-27
WIP: Add Query Optimizer that selects fields from query AST

NathHorrigan:
This is the initial code for optimizing GraphQL queries to reduce the number of database calls. The gist of this PR is that it analyses from the queries' AST what fields of what models it needs and then appends that to the Queryset object, further along the line when a `.specific()` call is applied these field selects are called on the Queryset using `.only(..)`, `.select_related(..)` & `.prefetch_related(..)`.

More accurate data will follow...

### DIFF
--- a/example/example/settings/base.py
+++ b/example/example/settings/base.py
@@ -54,6 +54,7 @@ INSTALLED_APPS = [
     "grapple",
     "graphene_django",
     "channels",
+    "django_extensions",
 ]
 
 MIDDLEWARE = [
@@ -178,3 +179,7 @@ CHANNEL_LAYERS = {
         "ROUTING": "grapple.urls.channel_routing",
     }
 }
+
+# Query Optimisation helpers
+SHELL_PLUS_PRINT_SQL = True
+RUNSERVER_PLUS_PRINT_SQL_TRUNCATE = 100000
--- a/grapple/utils.py
+++ b/grapple/utils.py
@@ -1,15 +1,18 @@
 import os
 import base64
 import tempfile
+import graphene_django_optimizer as gql_optimizer
 from PIL import Image, ImageFilter
 from django.conf import settings
 from wagtail.search.index import class_is_indexed
 from wagtail.search.models import Query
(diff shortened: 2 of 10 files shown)


NathHorrigan:
What do you think @zerolab??

example/example/settings/base.py
```diff
@@ -178,3 +179,7 @@
         "ROUTING": "grapple.urls.channel_routing",
     }
 }
+
+# Query Optimisation helpers
+SHELL_PLUS_PRINT_SQL = True
```
zerolab: these should be enabled only locally, maybe?

requirements.txt
```diff
@@ -10,3 +10,4 @@ factory-boy==2.12.0
 wagtail-factories==2.0.0
 django-cors-headers==3.0.2
 wagtail-headless-preview==0.1.4
+django_extensions==2.2.9
```
zerolab: I am not convinced we need the extra dependency. Should only be a dev dependency

grapple/utils.py
```diff
@@ -30,9 +32,10 @@ def resolve_queryset(
     :type order: str
     """
     offset = int(offset or 0)
+    qs = QueryOptimzer.query(qs, info)
 
     if id is not None:
-        qs = qs.filter(pk=id)
+        qs = qs.get(id=id)
```
zerolab: This can throw a `DoesNotExist` should the provided id not be there anymore

fabienheureux:
Why not consider using https://github.com/tfoxy/graphene-django-optimizer ?

NathHorrigan:
Hi @fabienheureux , I did try DQO but it doesn't work with stream field and that's a big issue for us. Instead of hack DQO into Grapple, I decided to create our own way of doing it that can be fine tuned and expanded in the future.

zerolab:
@NathHorrigan any chance you can bring this to the finish line?

ruisaraiva19:
> @NathHorrigan any chance you can bring this to the finish line?

@NathHorrigan any news on this PR?

PR #59, opened 2020-03-31
Add `register_query_field` helper decorator

NathHorrigan:
This PR adds the ability to add different root fields to the schema so the following code allows this query:

```python
# models.py
@register_snippet
@register_query_field('advert') <--- This decorator here
class Advert(models.Model):
    url = models.URLField(null=True, blank=True)

    panels = [FieldPanel("url"), FieldPanel("text")]

    graphql_fields = [GraphQLString("url"), GraphQLString("text")]

    def __str__(self):
        return self.text
```

```graphql
# query
{
  adverts {
    url
    text
  }
}
```


grapple/helpers.py
```diff
@@ -34,3 +37,57 @@ def register_streamfield_block(cls):
 def register_graphql_schema(schema_cls):
     registry.schema.append(schema_cls)
     return schema_cls
+
+
+def register_query_field(field_name, plural_field_name=None, query_params=None):
```
zerolab, line 42: undocumented `plural_field_name`

docs/general-usage/decorators.rst
```diff
@@ -0,0 +1,67 @@
+Decorators
+==========
+
+Here are some useful decorators that allow you to expand your GraphQL schema further
+
+@register_query_field
+---------------------
+.. module:: grapple.helpers
+.. class:: register_query_field(field_name, plural_field_name, query_fields)
```
zerolab, line 9:
```suggestion
.. class:: register_query_field(field_name, plural_field_name=None, query_fields=None)
```

PR #62, opened 2020-04-07
Add non-nullable support to `register_query_field` decorator

### LINKED ISSUE #61: Add non-nullable support to `register_query_field` help decorator
# 🙋 Feature Request

## 🔦 Context
Recently the non-nullable support was added to the Grapple fields on #51.
Now with the `register_query_field` help decorator, a similar approach would be nice!

## 😯 Describe the feature
Adding the same non-nullable functionality to the `register_query_field` help decorator for consistency across the GraphQL schema.
3 variables would be enough to cover all scenarios:

Add required return type for single queries:
```python
@register_query_field(
  'advert',
  'adverts',
  required=True,  # required single queries -> Item!
  plural_required=True,  # required plural queries list -> [Item]!
  plural_item_required=True,  # required for plural queries list items -> [Item!]!
)
```

With these 3 variables, we can have these return types on the GraphQL schema:

```graphql
# @register_query_field('advert')
advert(id: Int): Advert
adverts(id: Int, ...): [Advert]

# @register_query_field('advert', required=True)
advert(id: Int): Advert!
adverts(id: Int, ...): [Advert]

# @register_query_field('advert', plural_required=True)
advert(id: Int): Advert
adverts(id: Int, ...): [Advert]!

# @register_query_field('advert', plural_item_required=True)
advert(id: Int): Advert
adverts(id: Int, ...): [Advert!]

# @register_query_field('advert', plural_required=True, plural_item_required=True)
advert(id: Int): Advert
adverts(id: Int, ...): [Advert!]!
```

## 💻 Examples

Current behavior:

```python
# models.py
@register_snippet
@register_query_field('advert') <--- This decorator here
class Advert(models.Model):
    url = models.URLField(null=True, blank=True)

    panels = [FieldPanel("url"), FieldPanel("text")]

    graphql_fields = [GraphQLString("url"), GraphQLString("text")]

    def __str__(self):
        return self.text
```
Adds both `advert` and `adverts` queries to the GraphQL schema.
The `advert` query has a schema definition like this by default:
```graphql
advert(id: Int): Advert
```
Where the return type `Advert` shows as possibly `null`. And it's the same for the `adverts` (plural) query:
```graphql
adverts(
  limit: PositiveInt
  offset: PositiveInt
  order: String
  searchQuery: String
  id: ID
): [Advert]
```
For the plural query, the GraphQL schema has a possibly `null` return type for both the list and the list items.

## 💁 Possible Solution
I already had a bit of time and made an improved `register_query_field` decorator on a personal project.

**I'm going to open a pull request to add this feature.**

ruisaraiva19:
fix #61 

Tasks list:
- [x] non-nullable support to the singular query return type
- [x] non-nullable support to the plural query return list type
- [x] non-nullable support to the plural query return list item type
- [x] add documentation
- [x] add tests


grapple/helpers.py
```diff
@@ -69,18 +76,40 @@ def resolve_singular(self, _, info, **kwargs):
             def resolve_plural(self, _, info, **kwargs):
                 qs = cls.objects
                 if issubclass(cls, Page):
-                    qs = cls.objects.live().public().order_by("-first_published_at")
+                    qs = (
+                        cls.objects.live()
+                        .public()
+                        .specific()
```
zerolab, line 82: I am curious as to the need for `.specific()` here and below. Would be good to document

ruisaraiva19: Now that you mention, I did a quick test locally and indeed we don't need `.specific()` here 🙂 .
I'm going to remove it.

ruisaraiva19: I just pushed one commit with the update.

ruisaraiva19:
> This looks like an improvement to me. Could use a quick test to confirm the resulting schema is as described.
> Plus the question on specific

@zerolab Tests are now added to check the non-null query fields.

PR #74, opened 2020-06-15
Turn Graphene link from markdown to restructured format

tbrlpld:
The link was showing up in raw Markdown, because the docs are compiled form restructured text. Fixed the format to use restructured text embedded link (interestingly, that line break is part of the markup...).


zerolab:
Thanks @tbrlpld!

PR #77, opened 2020-06-25
Add clarifications on settings and functionality

tbrlpld:
Before it was not mentioned that Wagtail Headless Preview needs to be added 
to the list of installed apps or that it comes with migrations that need to be run. 
While this becomes clear when reading its docs, it is more convenient to include 
it here. 

Also, I was surprised that the tab that opens when I click on the "Preview" button is 
not opening actually the defined URL. Because I did not have an actual preview site 
in place at this time, I only got a default "page not found" from my browser. I think
the added explanation should should help other new users to not get stuck on that.

### DIFF
--- a/docs/general-usage/preview.rst
+++ b/docs/general-usage/preview.rst
@@ -14,8 +14,8 @@ support to any client whether that be a SPA or Native App.
 Setup
 ^^^^^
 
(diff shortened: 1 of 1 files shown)


zerolab:
Most useful additions @tbrlpld! LGTM

PR #89, opened 2020-08-05
Django 3.0 compatibility via optional channels

### LINKED ISSUE #73: Django 3 Compatiblity
Hey there! I'm wondering if there's plans to make this compatible with Django 3 and ASGI?

zerolab:
Attempt to make channels optional to pave the way for Django 3 support

The default `pip install wagtail_grapple` comes without Channels requirements.  To ge it you now need to run `pip install wagtail-grapple[channels]`

Fixes #73

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -39,7 +39,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v1
+        uses: actions/setup-python@v2
         with:
           python-version: 3.8
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -32,3 +32,19 @@ jobs:
           rm db.sqlite3
           python manage.py migrate
           python manage.py test
+
+
+  test_channels:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+
+      - name: Install Dependencies
+        run: |
+          pip install -r requirements-channels.txt
+          pip install .
+
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -47,4 +47,3 @@ jobs:
         run: |
           pip install -r requirements-channels.txt
           pip install .
-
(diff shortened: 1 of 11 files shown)


IAmNatch:
@zerolab, I'm happy to help test. Let me know what areas I should focus on, or if there's a process you aim for and I'll put in some time over the weekend.

zerolab:
@IAmNatch I'd try it on a test Django 3.0 project and see if there any issues.
I tested it with the example project in the repo, but getting more eyes would help

IAmNatch:
@zerolab I'm getting the following error when testing graphiql with this branch. 

```
File "/Users/shanerobbins/Development/personal/wagtail-grapple/grapple/types/search.py", line 3, in <module>
    from wagtail.documents.models import get_document_model
ImportError: cannot import name 'get_document_model' from 'wagtail.documents.models'
```

From what I'm seeing [here](https://github.com/wagtail/wagtail/commit/21ffe27db76392946dd95e43459eaebb162b790f), this function has been removed from `wagtail.documents.model`. It is still accessible from `wagtail.documents`, so this should be an easy fix.

zerolab:
Gah, I need to revise on latest master where I fixed that. If you do that locally you should be able to test

IAmNatch:
I made the change as suggested and was able to keep testing. Almost everything seemed to be running smoothly. No GraphQL errors. The only type of problem I ran into was mostly unrelated example folder updates.

Two issues I ran into in the example folder specifically:
1. Custom images did not work with Wagtail in the current version of the example. I would get an error trying to save the post with a valid image: `Select a
valid choice. That choice is not one of the available choices` I didn't see a clear solution to that so I just disabled the custom model setting and everything worked as expected.
2. The wagtail site middleware has been removed. https://docs.wagtail.io/en/latest/releases/2.9.html?highlight=sitemiddleware#sitemiddleware-and-request-site-deprecated 

I hope this is helpful!

.github/workflows/tests.yml
```diff
@@ -32,3 +32,18 @@ jobs:
...
           python manage.py migrate
           python manage.py test
+
+
+  test_channels:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
```
ruisaraiva19, line 42: You can update this action to v2
```yml
uses: actions/setup-python@v2
```

zerolab: I somehow missed they released v2. Updated

PR #90, opened 2020-08-09
Fix StructBlock/RichText output

NathHorrigan:
## In this PR I have:
- Fixed blocks method of `StructBlock` to allow use of RichText values
- Instead of returning the raw `RichText` value instead we pass it through the `wagtailcore/richtext.html` to allow customisation of the output.
- Added a default  `wagtailcore/richtext.html` template to remove default enclosing div (`<div class="rich-text">...</div>`)

Example Query:
Screenshot 2020-08-09 at 10 04 50


grapple/types/streamfield.py
```diff
@@ -146,15 +148,15 @@ def resolve_blocks(self, info, **kwargs):
             stream_data = self.value
             child_blocks = self.block.child_blocks
 
-        for field in stream_data:
-            block = child_blocks[field["type"]]
-            value = field["value"]
+        for (field, value) in stream_data.items():
```
zerolab, line 151:
```suggestion
        for field, value in stream_data.items():
```
should suffice

NathHorrigan:
Thanks Dan! Merging

PR #95, opened 2020-08-17
Add `url` to `DocumentObjectType`

### LINKED ISSUE #94: Add `url` field to `DocumentObjectType`
I know you guys probably pretty busy, but I just wanted to drop this feature request for future updates 😉

Currently the `DocumentObjectType` provides the following fields:
```
id: ID
title: String
file: String
createdAt: DateTime
fileSize: Int
fileHash: String
```

The `file` field resolves to some path like string like `"documents/example.pdf"`. Now to actually generate a link where I can access the document I need to pull the Wagtail GraphQL enpoint from the plugin options, trim it to the base URL (domain) and then add the correct media URL and then add the value from the `file` field. 

While this certainly works, it seems fragile and introduces some coupling between the frontend and backend. Anytime the backend might have a different media URL I would need to update that in the frontend (code or configuration). 

The `ImageObjectType` provides similar fields as the `DocumentObjectType` and many other fileds. One of them is the `src` field. This is a valid URL under which the respective image can be accessed. It would be great to add the same filed to the `DocumentObjectType` 😊

Since the document model already provides a `url` proptery just like the image model, I guess the [same approach](https://github.com/GrappleGQL/wagtail-grapple/blob/569b05d73282bee41b6c0509d413638b1aa9b040/grapple/types/images.py#L19) as for the `ImageObjectType.src` could be used for the `DocumentObjectType.src`.

tbrlpld:
As described in #94, it is currently necessary to construct the URL where a document can be loaded from the backend on the client side.
This is error prone and creates unnecessary coupling between the frontend and backend configuration (or code). 

The `ImageObjectType` on the other hand already provides a `src` field which will return a fully qualified URL under which the image can be reached. 

This PR implements the `src` field on the `DocumentObjectType`. The implementation is pretty much copied over from the `BaseImageObjectType`. 

A small set of tests have been added to ensure the functionality of the new field. Because I could not find other tests that directly test the queries for `images` and `documents` I have added the tests in `example/tests/test_grapple.py`.  A test for the `src` field on the `ImageObjectType` has been added in the same file for reference. 

Let me know what you think, and if any adjustments are necessary.

Closes #94

### DIFF
--- a/example/example/settings/base.py
+++ b/example/example/settings/base.py
@@ -154,6 +154,7 @@ MEDIA_URL = "/media/"
 
 WAGTAIL_SITE_NAME = "example"
 WAGTAILIMAGES_IMAGE_MODEL = "images.CustomImage"
+WAGTAILDOCS_SERVE_METHOD = "serve_view"
 
 # Base URL to use when referring to full URLs within the Wagtail admin backend -
 # e.g. in notification emails. Don't include '/admin' or a trailing slash
--- a/example/example/tests/test_grapple.py
+++ b/example/example/tests/test_grapple.py
@@ -94,7 +94,7 @@ class ImagesTest(BaseGrappleTest):
         self.assertEqual(example_img.id, 1)
         self.assertEqual(example_img.title, "Example Image")
 
-    def test_query_image_src(self):
+    def test_query_src_field(self):
         query = """
         {
             images {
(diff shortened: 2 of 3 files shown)


IAmNatch:
@zerolab Should media be updated to follow the same convention? Otherwise we've got 3 different API's between document, media and image

zerolab:
@IAmNatch sorry, should have been more explicit - Media should follow the same convention:

Attribute | Description
---------|-----------
`file`      | relative path
`url`     | full URL
`src`    | Legacy attribute on `Image`, deprecated (in a few releases time)

tbrlpld:
@zerolab Replying to https://github.com/GrappleGQL/wagtail-grapple/pull/96#issuecomment-674950403 

Sure! I will update the field name to `src` in the `DocumentObjectType`.

tbrlpld:
@zerolab Field name is change as of 41971e0. Tests have been updated accordingly (8a62053).

zerolab:
Awesome work, thank you @tbrlpld!

tbrlpld:
Happy to pitch in for this great project 👍

PR #96, opened 2020-08-17
Match document file handling to media file handling

### LINKED ISSUE #94: Add `url` field to `DocumentObjectType`

IAmNatch:
Fixes #94.

Currently the these graphQL types return the following values:
**Image:**
   file: `original_images/_myfile.png`
   src: `http://localhost:8000/media/original_images/_myfile.png`
**Media**: 
   file: `http://localhost:8000/media/media/myfile.MP4`
**Document**:
   file: `documents/myfile.pdf`

It was a bit difficult to determine the correct course of action, given the contradictions between the existing api. I made the assumption that images should remain as they are and that documents should match medias handling, since media has a file but no source. 

These proposed changes update the file resolver of document to match medias implementation, which looks like this:

**Document**
    file: `http://localhost:8000/media/documents/myfile.pdf`

If the maintainers of the project would prefer to add src to document and media I'd be happy to make that adjustment instead.

As well please note that in the same manner that media resolved its file, I utilize `self.file.url`. Another url could have been used via `self.url` which would have yielded `http://localhost:8000/documents/1/my.pdf`.  It was unclear which one would be ideal, so I deferred to the the media resolver's example. I did notice that the image object tries both...

### DIFF
--- a/grapple/types/documents.py
+++ b/grapple/types/documents.py
@@ -1,5 +1,6 @@
 import graphene
 
+from django.conf import settings
 from graphene_django.types import DjangoObjectType
 
 from wagtail import VERSION as WAGTAIL_VERSION
@@ -31,6 +32,11 @@ class DocumentObjectType(DjangoObjectType):
     created_at = graphene.DateTime(required=True)
     file_size = graphene.Int()
     file_hash = graphene.String()
+
+    def resolve_file(self, info, **kwargs):
+        if self.file.url[0] == "/":
+            return settings.BASE_URL + self.file.url
+        return self.file.url
 
 
 def DocumentsQuery():


tbrlpld:
Wow. Almost at the same time 😅 #95

IAmNatch:
Nice work @tbrlpld! Your implementation looks very fleshed out in terms of testing so I'm happy for them to pull of yours. Lets just confirm with @zerolab whether the URL should be accessible via `src` or via `file`. 

Either way I'm glad this functionality is being consolidated across media, document and image.

zerolab:
Let's go with `url` as the full URL, and keep file as the partial. `src` assumes that it'll be used as the value of a particular attribute, rather than describing what the value is.

@tbrlpld your PR is closest, are you happy to update that? 
Leave `src` on Image with a TODO: to remove in the future. Other than that #95 LGTM.

Thank you for this @IAmNatch.

PR #97, opened 2020-08-18
Use consistent media url attributes

zerolab:
Follow up to #95

### DIFF
--- a/grapple/types/images.py
+++ b/grapple/types/images.py
@@ -1,9 +1,5 @@
-import os
-import codecs
-import urllib.parse
 import graphene
 
-from django.conf import settings
 from graphene_django import DjangoObjectType
 from wagtail.images import get_image_model
 from wagtail.images.models import (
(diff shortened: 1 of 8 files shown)


zerolab:
@tbrlpld this is the follow up to your PR to make everything consintent

grapple/types/images.py
```diff
@@ -12,34 +8,29 @@
...
+    def resolve_url(self, info):
+        """
+        Get the uploaded image url.
+        """
+        return get_media_item_url(self)
+
     def resolve_src(self, info):
         """
-        Get url of the original uploaded image.
+        Deprecated. Use the `url` attribute.
```
ruisaraiva19, line 31: This deprecation can be added to the GraphQL schema through the `deprecation_reason` scalar argument.
```python
src = graphene.String(required=True, deprecation_reason="Use the `url` attribute")
```

ruisaraiva19 (changes requested):

grapple/types/images.py
zerolab: I knew I did not search well enought. Kudos for the tip!

ruisaraiva19 (approved):

PR #106, opened 2020-09-02
Make wagtailmedia an optional dependency

zerolab:
wagtailmedia is required, but not documented.
It is best to keep this optional.

To-Do:

- [ ]  Better handling of optionality. Maybe add as separate app?
- [ ] Tests

### DIFF
--- a/grapple/actions.py
+++ b/grapple/actions.py
@@ -9,7 +9,6 @@ from django.conf import settings
 from django.template.loader import render_to_string
 
 from graphene_django.types import DjangoObjectType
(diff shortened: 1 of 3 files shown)


ruisaraiva19 (commented):
Looking good so far 🙂.

zerolab:
@ruisaraiva19 @NathHorrigan I plan to make the `Query` definition extensible (via hooks), so once that is done can move wagtailmedia support into sub-app and write proper tests since at the moment the example app required wagtailmedia.

As such I am merging this

zerolab:
Wider follow-up - #126

PR #108, opened 2020-09-04
Add token/slug support to register_query_field/register_paginated_query_field

### LINKED ISSUE #107: Handle preview tokens with register_query_field / register_paginated_query_field
This is handled by the generic `page` query via `get_specific_page()` - 
https://github.com/GrappleGQL/wagtail-grapple/blob/5249f670f527db2735ad4cf5634e6bfbed582f60/grapple/types/pages.py#L150

zerolab:
Fixes #107

### DIFF
--- a/grapple/helpers.py
+++ b/grapple/helpers.py
@@ -55,7 +55,11 @@ def register_query_field(
 
     def inner(cls):
         field_type = lambda: registry.models[cls]
-        field_query_params = query_params or {"id": graphene.Int()}
+        field_query_params = query_params or {
+            "id": graphene.Int(),
+            "slug": graphene.String(),
+            "token": graphene.String(),
+        }
 
         def Mixin():
             # Generic methods to get all and query one model instance.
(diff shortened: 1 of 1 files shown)


grapple/helpers.py
```diff
@@ -64,6 +68,9 @@ def resolve_singular(self, _, info, **kwargs):
                 if not kwargs:
                     return None
 
+                if "token" in kwargs and hasattr(cls, "get_page_from_preview_token"):
```
ruisaraiva19, line 71: Should we move this logic inside `if issubclass(cls, Page):` to only call `get_page_from_preview_token` on pages?

grapple/helpers.py
```diff
@@ -145,6 +156,9 @@ def resolve_singular(self, _, info, **kwargs):
...
```
ruisaraiva19, line 159: Same here.

grapple/helpers.py
zerolab: good shout!

ruisaraiva19 (approved):

PR #109, opened 2020-09-08
Use url instead of urlPath in redirections

fabienheureux:
When using redirects, let's assume you have your homepage's slug set on `/home` if you add redirections from `/blog` to your homepage (or any homepage's children), the redirection will be `/home/[slug]` instead of `/[slug]`

This fixes it.


zerolab:
Thank you @fabienheureux

PR #111, opened 2020-09-08
Add collections types

fabienheureux:
[Wagtail's collections](https://docs.wagtail.io/en/v2.8.1/editor_manual/documents_images_snippets/collections.html) are pretty useful, and I noticed that they were not retrieved by wagtail-grapple by default. 

This is a bit of a work in progress, I took inspiration from `grapple.types.images`, but maybe I got something wrong

This PR could use some help, especially
- [ ] Add tests

Let me know where it can be improved 😉

### DIFF
--- a/grapple/actions.py
+++ b/grapple/actions.py
@@ -90,7 +90,8 @@ def add_app(app_label: str, prefix: str = ""):
 
     # Add snippet models to model collection.
     for snippet in get_snippet_models():
-        models.append(snippet)
+        if snippet._meta.app_label == app_label:
+            models.append(snippet)
 
     # Create add each model to correct section of registry.
     for model in models:
--- a/grapple/actions.py
+++ b/grapple/actions.py
@@ -90,8 +90,7 @@ def add_app(app_label: str, prefix: str = ""):
 
     # Add snippet models to model collection.
     for snippet in get_snippet_models():
-        if snippet._meta.app_label == app_label:
-            models.append(snippet)
+        models.append(snippet)
 
     # Create add each model to correct section of registry.
     for model in models:
--- a/grapple/types/collections.py
+++ b/grapple/types/collections.py
@@ -23,7 +23,7 @@ class CollectionObjectType(DjangoObjectType):
     def resolve_descendants(self, info, **kwargs):
         return self.get_descendants()
 
-    def resolve_ancestors(self, info):
+    def resolve_ancestors(self, info, **kwargs):
         return self.get_ancestors()
 
 
@@ -32,8 +32,8 @@ def CollectionsQuery():
     mdl_type = registry.collections.get(mdl, CollectionObjectType)
 
     class Mixin:
-        collections = QuerySetList(mdl_type, enable_search=False)
-        collection_type = graphene.String()
+        collections = QuerySetList(mdl_type, enable_search=False, required=True)
+        collection_type = graphene.String(required=True)
 
         # Return all collections
         def resolve_collections(self, info, **kwargs):
--- a/grapple/types/collections.py
+++ b/grapple/types/collections.py
@@ -19,13 +19,6 @@ class CollectionObjectType(DjangoObjectType):
     name = graphene.String()
     descendants = graphene.List(lambda: CollectionObjectType)
     ancestors = graphene.List(lambda: CollectionObjectType)
-    contents = graphene.List(lambda: CollectionObjectType)
-
-    def resolve_contents(self, info):
-        return list([
-            hook(self)
-            for hook in hooks.get_hooks("describe_collection_contents")
-        ])
 
     def resolve_descendants(self, info):
         return self.get_descendants()
--- a/grapple/types/collections.py
+++ b/grapple/types/collections.py
@@ -15,12 +15,12 @@ class CollectionObjectType(DjangoObjectType):
     class Meta:
         model = Collection
 
-    id = graphene.ID()
-    name = graphene.String()
-    descendants = graphene.List(lambda: CollectionObjectType)
-    ancestors = graphene.List(lambda: CollectionObjectType)
+    id = graphene.ID(required=True)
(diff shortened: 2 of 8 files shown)


grapple/types/images.py
```diff
@@ -10,15 +10,20 @@
 from ..registry import registry
 from ..utils import resolve_queryset, get_media_item_url
 from .structures import QuerySetList
+from .collections import CollectionObjectType
 
 
 class BaseImageObjectType(graphene.ObjectType):
+    src = graphene.String()
+    aspect_ratio = graphene.Float()
+    sizes = graphene.String()
```
ruisaraiva19, line 19: This is duplicated, line 25 already has this field declared.

grapple/types/images.py
```diff
@@ -72,6 +77,7 @@ class ImageObjectType(DjangoObjectType, BaseImageObjectType):
         jpegquality=graphene.Int(),
     )
     src_set = graphene.String(sizes=graphene.List(graphene.Int))
+    tags = graphene.List(graphene.String)
```
ruisaraiva19, line 80: Not sure if we want to handle the tags on this PR.
Since tags are used on other models, they should have the same logic.
@NathHorrigan @zerolab what do you think of adding `tags` fields and resolvers on a new PR?

grapple/types/redirects.py
```diff
@@ -23,7 +23,7 @@ def resolve_new_url(self, info, **kwargs):
         if self.redirect_page is None:
             return self.link
 
-        return self.redirect_page.url_path
+        return self.redirect_page.url
```
ruisaraiva19, line 26: This was already updated on #109. Please merge the latest changes from the `master` branch.

grapple/actions.py
```diff
@@ -72,8 +72,7 @@ def add_app(app_label: str, prefix: str = ""):
 
     # Add snippet models to model collection.
     for snippet in get_snippet_models():
-        if snippet._meta.app_label == app_label:
-            models.append(snippet)
+        models.append(snippet)
```
ruisaraiva19, line 75: I'm not sure is this will throw any errors 🤔.

ruisaraiva19 (changes requested):

grapple/types/images.py
fabienheureux: This was not meant to be committed. This was a WIP for another PR...

grapple/actions.py
fabienheureux: Hey, first thanks a lot for your feedbacks, much appreciated ! 

I am not sure what was the reason of this change, I should have added some comments...but I think that when "collection" are added registered as snippet, the check on the app_label would not add snippet to model collection. 

I will start working on it next week, I currently use this branch as is on production without issues, but it may not cover all cases of snippet's usage 🤔

grapple/types/images.py
```diff
@@ -19,6 +20,7 @@ class BaseImageObjectType(graphene.ObjectType):
     url = graphene.String(required=True)
     aspect_ratio = graphene.Float(required=True)
     sizes = graphene.String(required=True)
+    collection = graphene.Field(lambda: CollectionObjectType)
```
ruisaraiva19, line 23: This should be non-nullable as well.
```python
collection = graphene.Field(lambda: CollectionObjectType, required=True)
```

ruisaraiva19:
@fabienheureux Only a couple of tweaks needed and should be good to go!

fabienheureux:
@ruisaraiva19 thanks for the review, I just add updates to take your latest feedbacks into account.

Let me know if this can be improved.

grapple/types/collections.py
```diff
@@ -0,0 +1,45 @@
+from wagtail.core.models import Collection
+from wagtail.core import hooks
```
zerolab, line 2: unused import.

the order should be:
```python
import graphene

from graphene_django.types import DjangoObjectType 
from wagtail.core.models import Collection

from ..registry import registry
from ..utils import resolve_queryset
from .structures import QuerySetList
```

grapple/types/images.py
```diff
@@ -19,6 +20,7 @@ class BaseImageObjectType(graphene.ObjectType):
     url = graphene.String(required=True)
     aspect_ratio = graphene.Float(required=True)
     sizes = graphene.String(required=True)
+    collection = graphene.Field(lambda: CollectionObjectType, required=True)
```
zerolab, line 23: the collection field should be added to `DocumentObjectType` too

fabienheureux:
Hey, thanks for the review as well @zerolab , I just updated the PR. 
Also, I updated the PR to be compatible with recent changes made to master branch, do you confirm that I got it right ? https://github.com/fabienheureux/wagtail-grapple/commit/8e5062438160745bc20b0a60acbc9e335180caa9 

Thanks !

zerolab:
Thank you @fabienheureux 
Will merge this tomorrow and follow up with a few changes, mainly to only return items in public collections, as well as only return public collection via the collection query

We do the same for Pages and since we are not yet doing mutations, public data it is

fabienheureux:
Ok perfect ! Thanks for the follow up. 

Not directly related, but would you be interested in adding images tags to wagtail-grapple, or is it out of the package's scope ?
I had a PR related to this specific addition, and if this is something you would consider merging, I will try to finish it in the following weeks.

zerolab:
> would you be interested in adding images tags to wagtail-grapple
Absolutely

zerolab:
Thank you @fabienheureux.
Added the limiting in 60fa9a5 and dropped `collection_type` from the query mixin in 2f2f4fa8.

cc @ruisaraiva19

fabienheureux:
> > would you be interested in adding images tags to wagtail-grapple
> 
> Absolutely!

Ok ! I will start working on this in following weeks.
Thanks for the reviews and merge !

PR #112, opened 2020-09-08
register_singular_query_field decorator

zerolab:
This PR adds a `@register_singular_query_field` decorator that allows one to pull the first item of a particular type.
This is especially useful when you have a Wagtail Page with `max_num = 1`

e.g. 

```python
from wagtail.core.models import Page


@register_singular_query_field("the_page")
class ThePage(Page):
    max_count = 1
```

allows you to query:

```graphql
{
   thePage {
        title
   }
}
```

to get your page, without having to worry about its id

### DIFF
--- a/grapple/helpers.py
+++ b/grapple/helpers.py
@@ -218,15 +218,21 @@ def register_paginated_query_field(
 def register_singular_query_field(field_name, query_params=None, required=False):
     def inner(cls):
         field_type = lambda: registry.models[cls]
-        field_query_params = query_params or {
-            "token": graphene.Argument(
-                graphene.String, description=ugettext_lazy("The preview token.")
-            ),
-            "order": graphene.Argument(
-                graphene.String,
-                description=ugettext_lazy("Use the Django QuerySet order_by format."),
-            ),
-        }
+        field_query_params = query_params
+
+        if field_query_params is None:
+            field_query_params = {
+                "order": graphene.Argument(
+                    graphene.String,
+                    description=ugettext_lazy(
+                        "Use the Django QuerySet order_by format."
+                    ),
+                ),
+            }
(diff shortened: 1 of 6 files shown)


grapple/helpers.py
```diff
@@ -217,3 +213,61 @@ def resolve_plural(self, _, info, **kwargs):
...
+                            cls, "get_page_from_preview_token"
+                        ):
+                            return cls.get_page_from_preview_token(kwargs["token"])
+
+                        qs = cls.objects.live().public()
+                        if "order" in kwargs:
+                            qs = qs.order_by(kwargs.pop("order"))
+                        return qs.filter(**kwargs).first()
+
+                    return cls.objects.first()
```
ruisaraiva19, line 247: The `order` argument isn't being used here.
Updating to `return cls.objects.filter(**kwargs).first()` should do the trick.

ruisaraiva19 (changes requested):

grapple/helpers.py
zerolab: nice catch. sorted!

also changed to only add the default token param for `Page` subclasses in d28fa5b

ruisaraiva19 (approved):

grapple/helpers.py
```diff
@@ -51,11 +51,17 @@ def register_query_field(
 
     def inner(cls):
         field_type = lambda: registry.models[cls]
-        field_query_params = query_params or {
```
zerolab, line 53: @ruisaraiva19 changed the other two decorators to add the default slug/token params only for page subclasses

ruisaraiva19 (approved):

grapple/helpers.py
ruisaraiva19: nice!

PR #115, opened 2020-09-20
Multi-site support - add SitesQuery and resolve pages based on site

### LINKED ISSUE #114: Multi-site support
**Issue**
It would be nice if `wagtail-grapple` would support a multi-site setup. E.g., there currently is no support for filtering page/pages based on a site. In a multi-site scenario, duplicate slugs can exist, and fetching a page would result in no results from the page query.

**My idea**
Add a filter to the page/pages query. We could for example introduce `hostname` as an optional parameter for these queries and filter the page QuerySet that way.

**Open questions**

- Searching; how do we handle searching on pages for a site?
- ?

leewesleyv:
Resolves #114

### DIFF
--- a/grapple/types/pages.py
+++ b/grapple/types/pages.py
@@ -189,7 +189,10 @@ def PagesQuery():
 
     class Mixin:
         pages = QuerySetList(
-            graphene.NonNull(lambda: PageInterface), enable_search=True, required=True
+            graphene.NonNull(lambda: PageInterface),
+            in_site=graphene.Boolean(),
+            enable_search=True,
+            required=True,
         )
         page = graphene.Field(
             PageInterface,
@@ -197,16 +200,18 @@ def PagesQuery():
             slug=graphene.String(),
             token=graphene.String(),
             content_type=graphene.String(),
+            in_site=graphene.Boolean(),
         )
 
         # Return all pages in site, ideally specific.
         def resolve_pages(self, info, **kwargs):
-            site = Site.find_for_request(info.context)
-            return resolve_queryset(
-                WagtailPage.objects.in_site(site).live().public().specific(),
-                info,
-                **kwargs
-            )
+            pages = WagtailPage.objects.live().public().specific()
+
+            if kwargs.get("in_site", False):
+                site = Site.find_for_request(info.context)
+                pages = pages.in_site(site)
+
+            return resolve_queryset(pages, info, **kwargs)
 
         # Return a specific page, identified by ID or Slug.
         def resolve_page(self, info, **kwargs):
@@ -215,7 +220,9 @@ def PagesQuery():
                 slug=kwargs.get("slug"),
                 token=kwargs.get("token"),
                 content_type=kwargs.get("content_type"),
-                site=Site.find_for_request(info.context),
+                site=Site.find_for_request(info.context)
+                if kwargs.get("in_site", False)
+                else None,
             )
 
     return Mixin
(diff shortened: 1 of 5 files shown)


grapple/types/pages.py
```diff
@@ -191,10 +196,13 @@ class Mixin:
             content_type=graphene.String(),
         )
 
-        # Return all pages, ideally specific.
+        # Return all pages in site, ideally specific.
         def resolve_pages(self, info, **kwargs):
+            site = Site.find_for_request(info.context)
             return resolve_queryset(
-                WagtailPage.objects.live().public().specific(), info, **kwargs
+                WagtailPage.objects.in_site(site).live().public().specific(),
```
ruisaraiva19, line 203: This is a breaking change.
We may add an argument like `in_site=graphene.Boolean()` and when `True` we would filter pages with `.in_site(site)`.
This way, we add the feature and keep backward compatibility 🙂.

grapple/types/pages.py
```diff
@@ -204,6 +212,7 @@ def resolve_page(self, info, **kwargs):
                 slug=kwargs.get("slug"),
                 token=kwargs.get("token"),
                 content_type=kwargs.get("content_type"),
+                site=Site.find_for_request(info.context),
```
ruisaraiva19, line 215: Same here regarding backward compatibility.
```python
site=Site.find_for_request(info.context) if kwargs.get("in_site", False) else None,
```

grapple/types/sites.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+
+    class Meta:
+        model = Site
+
+
+def SitesQuery():
+    class Mixin:
+        site = graphene.Field(
+            SiteType, hostname=graphene.String(), id=graphene.String()
```
ruisaraiva19, line 44: Please update the id argument to be `graphene.ID()` instead of `graphene.String()`.

grapple/types/sites.py
```diff
@@ -0,0 +1,66 @@
...
+        )
+        sites = QuerySetList(
+            graphene.NonNull(SiteType), enable_search=True, required=True
+        )
+
+        # Return all sites.
+        def resolve_sites(self, info, **kwargs):
+            return resolve_queryset(Site.objects.all(), info, **kwargs)
+
+        # Return all sites, identified by ID or hostname.
```
ruisaraiva19, line 54: Please update this copy to something like:
```diff
-        # Return all sites, identified by ID or hostname.
+        # Return a site, identified by ID or hostname.
```

ruisaraiva19 (changes requested):

ruisaraiva19:
@leewesleyv I left a couple of comments.
Let me know if you have any questions.
Also, there is one file with merge conflicts to take care of 🙂.

grapple/types/pages.py
zerolab: This was based on my note from https://github.com/GrappleGQL/wagtail-grapple/issues/114#issuecomment-694905294

The REST API uses `in_site` as a filter. I have since seen a few queries about multi-site via a single endpoint which makes me think the site filtering should be opt-in. So good shout @ruisaraiva19 !

zerolab:
@leewesleyv excellent work on this! Would be great if you could make the `in_site` logic conditional and off by default.
And ideally add some documentation in `docs/general-usage/graphql-types.rst` (https://wagtail-grapple.readthedocs.io/en/latest/general-usage/graphql-types.html)

leewesleyv:
@ruisaraiva19 and @zerolab thanks for the comments! I will make sure that the code stays backward compatible and resolve the other issues. I will also update the documentation regarding the new functionality.

zerolab (approved):

zerolab:
Thank you for seeing this through @leewesleyv

PR #119, opened 2020-10-04
Add GraphQLPage to model types in docs

### LINKED ISSUE #116: GraphQLForeignKey doesn't work with "wagtailcore.Page" for content_type
trying to add `GraphQLForeignKey` pointing to `wagtailcore.Page` results in `Could not import 'grapple.schema.schema' for Graphene setting 'SCHEMA'. AttributeError: 'NoneType' object has no attribute '_meta'.`

Use case:
- an orderable with a page chooser that allows any type of page
- you want a list of chosen pages in the API

As far as I can see, the issue is around
https://github.com/GrappleGQL/wagtail-grapple/blob/e84622a8c67a0a684a59c1ba0d3448baf541a305/grapple/models.py#L81-L87 and the fact that the core page is not added to models. instead it is added as a `DjangoObjectType`

zerolab:
Fixes #116

### DIFF
--- a/docs/general-usage/model-types.rst
+++ b/docs/general-usage/model-types.rst
@@ -307,8 +307,8 @@ GraphQLForeignKey
 .. class:: GraphQLForeignKey(field_name, content_type, required=False)
 
     ``GraphQLForeignKey`` is similar to ``GraphQLSnippet`` in that you pass a
-    ``field_name`` and ``content_type`` but you can also specify that the field
-    is a list (for example when using ``Orderable``).
+    ``field_name`` and ``content_type``. You can also specify that the field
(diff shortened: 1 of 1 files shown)


ruisaraiva19 (approved):

tbrlpld:
I think this looks good! 

One question would be if the `is_list` argument should be documented with the other useful keyword arguments. If using the `GraphQLCollection` is the preferred way over the `is_list` flag, then can leave it out of course.

zerolab:
is_list, required, source all come from `GraphQLField` which is the base class for all the other. that one needs documenting instead. will open separate issue

tbrlpld:
Yea that sounds good. I was wondering how I even found the `is_list` argument ^^

Issue #128, opened 2020-10-07
Private and unpublished Pages are shown in children


mmmente:
Hi,

I just realized, that unpublished and private pages are listed in children and via there can be accessed.
I assume the `resolve_children()` needs to get a `.live().public()` at the end. If accessing a page directly, they do not show up. 

Or is this expected behaviour?

zerolab:
`get_specific_page()` does do `.live().public()`, so does `resolve_pages` and considering that we are not yet doing mutations is should likely be public everywhere. (i.e. `resolve_parent`,  `resolve_children`, `resolve_siblings`, `resolve_next_siblings`, `resolve_previous_siblings`, `resolve_descendants` and `resolve_ancestors`

Care to open an PR with the changes?

mmmente:
Done, see #129 
Not sure why the test fails with the message of 403 to the repository...

zerolab:
fixed in #129

PR #129, opened 2020-10-07
Add .live().public() to all relevant methods

mmmente:
As described in #128 private and draft pages are included in some querysets. Fixed this by including` .live().public()` to all relevant querys

### DIFF
--- a/grapple/types/pages.py
+++ b/grapple/types/pages.py
@@ -89,7 +89,9 @@ class PageInterface(graphene.Interface):
         Resolves a list of live children of this page.
         Docs: https://docs.wagtail.io/en/stable/reference/pages/queryset_reference.html#examples
         """
-        return resolve_queryset(self.get_children().live().public().specific(), info, **kwargs)
+        return resolve_queryset(
+            self.get_children().live().public().specific(), info, **kwargs
+        )
 
     def resolve_siblings(self, info, **kwargs):
         """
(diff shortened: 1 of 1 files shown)


grapple/types/pages.py
```diff
@@ -80,7 +80,7 @@ def resolve_parent(self, info, **kwargs):
         Docs: https://docs.wagtail.io/en/stable/reference/pages/model_reference.html#wagtail.core.models.Page.get_parent
         """
         try:
-            return self.get_parent().specific
+            return self.get_parent().live().public().specific
```
ruisaraiva19, line 83: `self.get_parent()` returns a `Page` and `Page` doesn't have the `.live()` and `.public()` methods.

ruisaraiva19 (changes requested):

grapple/types/pages.py
mmmente: True, I missed that. Actually in that situation it makes sense to not check for live and public.

zerolab:
@mmmente you need run `black grapple` to lint (that is the CI failure).

grapple/types/pages.py
ruisaraiva19: Nice. Now just fix the linting issues and will be good to go.

zerolab:
@ruisaraiva19 once the linting is done, feel free to merge

mmmente:
@zerolab Linting done, thanks for the hint!
@ruisaraiva19 go ahead :)

ruisaraiva19 (approved):

ruisaraiva19:
> @zerolab Linting done, thanks for the hint!
> @ruisaraiva19 go ahead :)

@mmmente thank you!

PR #130, opened 2020-10-08
Add webpquality support for renditions

### LINKED ISSUE #125: Support webpquality option for renditions
https://docs.wagtail.io/en/stable/topics/images.html#image-quality

ruisaraiva19:
Resolves #125 

By default, all `bmp` and `webp` images are converted to the `png` format when no image output format is given.
The default conversion mapping can be changed ([docs here](https://docs.wagtail.io/en/stable/advanced_topics/images/image_file_formats.html#customizing-output-formats)).

```graphql
{
  image(id: 2) {
    id
    url
    rendition(webpquality: 75) {
      url
    }
  }
}
```

```yaml
{
  "data": {
    "image": {
      "id": 2,
      "url": "http://localhost:8000/media/original_images/file_example_WEBP_50kB.webp",
      "rendition": {
        "url": "/media/images/file_example_WEBP_50kB.webpquality-75.png"
      }
    }
  }
}
```

### DIFF
--- a/docs/general-usage/graphql-types.rst
+++ b/docs/general-usage/graphql-types.rst
@@ -130,6 +130,7 @@ need for Gatsby Image features to work (see Handy Fragments page for more info):
         format: String
         bgcolor: String
         jpegquality: Int
+        webpquality: Int
     ): ImageRenditionObjectType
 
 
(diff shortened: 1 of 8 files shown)


zerolab (approved):
LGTM.
Need to check how Wagtail tackles this, but I wonder what happens when you pass `webpquality` to a jpeg and viceversa 🤔

zerolab:
48d2e1b makes me think we are missing some tests - would be good to get an idea of the type of query/queries that made you discover the inconsistency. 

Nonetheless, let's get this in and follow up with ^

PR #132, opened 2020-10-19
Add hook for registering query mixins

### LINKED ISSUE #126: Make the Query extensible when defining Schema
e.g. to replace an existy query type (e.g search), or better handle optional queries like `MediaQuery` for wagtailmedia

zerolab:
Fixes #126

### DIFF
--- a/docs/general-usage/hooks.rst
+++ b/docs/general-usage/hooks.rst
@@ -1,0 +1,34 @@
+Hooks
+=====
+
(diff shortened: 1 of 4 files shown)


ruisaraiva19 (approved):
LGTM

zerolab:
Initially I hoped to move the wagtailmedia bits to own app, even toyed with the idea of moving this definition to wagtailmedia itself (which I may do), but this is as far as the hooks bit goes, so will merge

Issue #138, opened 2020-11-07
DOCS: Add reference to gatsby-source-wagtail in the usage guide


thclark:
Hi folks,

Steadily implementing grapple as a side project, and have spent a lot of time getting `gatsby-source-graphql` working to pull in wagtail stuff for a frontend build.

However, I've just found the [`gatsby-source-wagtail`](https://www.gatsbyjs.com/plugins/gatsby-source-wagtail/) plugin (which was referenced in issue #81)! I had searched pretty extensively around graph, wagtail and gatsby but not come across it before.

Although I understand the strict purpose of this repo is to provide a graphQL endpoint for wagtail (which is usable by far more than just gatsby consumers), I think that the main use case would be for gatsby builds.

I therefore think that the 'Usage' section of the docs (which presently focuses solely on the backend), would be complemented really nicely by small section like this (in restructuredtext for your copy-paste convenience):

```
Using Grapple with Gatsby
====================

``wagtail-grapple`` provides a graphQL API to wagtail's contents which can be accessed by a wide range of consumers. However, probably the most common use case is using Gatsby to fetch contents from the API then build and deploy a static site build in React.

Frontend developers using Gatsby should check out `gatsby-source-wagtail <https://www.gatsbyjs.com/plugins/gatsby-source-wagtail/>`_, a plugin designed to interface directly with wagtail-grapple and ease the pain of querying site data.

Additionally, the `wagtail-gatsby <https://github.com/NathHorrigan/wagtail-gatsby>`_ library provides a set of hooks allowing gatsby builds to be triggered when pages are updated.
```

zerolab:
Hey @thclark,

While I disagree with the statement that Grapple is used primarily with Gatsby, I see the benefit of adding references to other packages and frontend frameworks and more. Feel free to submit a PR

thclark:
@zerolab PR Made. I changed the wording to "One common use case" rather than "Probably the most common"

As an aside, I'd like to hear what your view of the primary purpose of grapple is, if not to enable use as a headless CMS?

zerolab:
Thank you.

My view on it is that it provides an API without a prescriptive end use. It so happens the current usage for GraphQL endpoints is predominantly in the headless space. And that is great. 

I would, however, steer away from equating headless with Gatsby, which was my main gripe. We use Grapple with Nuxt at the moment. Others with Next.js. At the end of the day, one should use the tool that fits the job, and my hope is that Grapple helps with that.

[...]

PR #139, opened 2020-11-07
Adds a "Usage with Gatsby" section to the docs

thclark:
Based on the proposal in #138


zerolab (changes requested):
Thank you for this. Already a marked improvement.

Can you please use `Wagtail` (uppercase W) and   `GraphQL` everywhere for consistency? Other than that LGTM.

We may end renaming that section when we add more frameworks for reference, but that is for the future.

PR #153, opened 2021-02-03
Fix: custom document model not working with wagtail grapple

fabienheureux:
Current version of Grapple does not seem to work with custom document model. 
This fixes it

- [x] Fix the issue
- [x] Add tests

Edit: I added a custom document model in the example project in order to ensure that a custom document model works with Grapple. 
Sorry for all the commits, this is the first time I worked with the example project and a few things went wrong because I did not install pre-commit hooks :sweat_smile: 

Also: I had to port this fix in order to fix example project tests: https://github.com/wagtail/wagtail/pull/6559/files 
Let me know if I should rather fix this in another PR.


example/home/migrations/0002_create_homepage.py
```diff
@@ -48,6 +48,9 @@ def remove_homepage(apps, schema_editor):
 
 
 class Migration(migrations.Migration):
+    run_before = [
+        ("wagtailcore", "0053_locale_model"),
```
fabienheureux, line 52: cf https://github.com/wagtail/wagtail/pull/6559/files

zerolab:
@fabienheureux another omission in tests! Can you add a test to confirm your custom doc model is picked up?

fabienheureux:
@zerolab I updated the code a bit to make the document type file to look closer to the image type file.   
It helped me debugging and it will make the codebase more consistent, what do you think ? 

And I added a test 😅 (sorry for the delay.....).

zerolab:
Thank you @fabienheureux, merged with a tweak

PR #154, opened 2021-02-09
PageInterface swap

Leden:
This gives users the ability to swap the implementation of `PageInterface`.

An example use-case is to override the resolvers for the fields, for example, to allow API access to private pages based on current user's permissions, etc.

Usage:

```
# settings.py
GRAPPLE_PAGE_INTERFACE = "myproject.myapp.interfaces.MyPageInterface"
```
If not set, uses the default implementation

```
# myproject/myapp/interfaces.py
from grapple.types.pages import PageInterface

class MyPageInterface(PageInterface):
    def resolve_children(self, into, **kwargs):
        # alternative implementation of the `children` field resolver
        return ...
```

### DIFF
--- a/grapple/actions.py
+++ b/grapple/actions.py
@@ -22,7 +22,7 @@ from wagtail.snippets.models import get_snippet_models
 from .registry import registry
 from .types.documents import DocumentObjectType
 from .types.images import ImageObjectType
-from .types.pages import PageInterface, Page
+from .types.pages import Page, get_page_interface
 from .types.streamfield import generate_streamfield_union
 from .helpers import streamfield_types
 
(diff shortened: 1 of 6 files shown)


zerolab:
Thank you for this @Leden.
Could you please add some documentation and a couple of tests?

dopry:
@Leden I wanted to check in and see if you had any time or interest in finishing up this PR?

dopry:
@zerolab how do we want to proceed on this PR? @Leden  doesn't appear to be responding. Do you think this is a valuable enough feature for one of us to take over or should we just close it as stale?

zerolab:
Considering that a project that does not use Grapple can implement any custom interface, this can be a valuable enough addition to allow folk to customise things.

dopry:
superceded by: #325

PR #159, opened 2021-03-07
Add ability to limit the renditions Grapple will generate

### LINKED ISSUE #158: Querying arbitrary renditions could be a vector for DoS attacks
I spotted this on a project recently and wasn’t quite sure what to make of it – in its current shape, Grapple’s [`ImageObjectType`](https://wagtail-grapple.readthedocs.io/en/latest/general-usage/graphql-types.html) supports querying arbitrary renditions:

```
rendition(
    max: String
    min: String
    width: Int
    height: Int
    fill: String
    format: String
    bgcolor: String
    jpegquality: Int
    webpquality: Int
): ImageRenditionObjectType
```

This works very well with the GraphQL way of defining page requirements client-side, compared to the django-rest-framework equivalent `ImageRenditionField`. In Wagtail, creating renditions is quite an onerous process though, which I believe could open the door to Denial-of-Service attacks – query a GraphQL endpoint with tens of different renditions, and there you have a response that will take seconds to generate (and start filling your file storage).

This is a problem for GraphQL endpoints that are publicly-available, so a Gatsby site for instance might not suffer from this, but something else that generates queries in the browser, would.

---

Is there anything that could be implemented in Grapple to mitigate this? Having a form of "allowed renditions" check in the Graphene schema / types perhaps? If not I think this should at least be documented as a known shortcoming, so implementers are more aware of the risk.

zerolab:
This PR fixes #158 by adding a new `GRAPPLE_ALLOWED_IMAGE_FILTERS` setting that is a list of allowed image filters (which constitute the parameters for a rendition) before returning a rendition or a srcSet

### DIFF
--- a/docs/general-usage/graphql-types.rst
+++ b/docs/general-usage/graphql-types.rst
@@ -166,6 +166,9 @@ GRAPPLE_ALLOWED_IMAGE_FILTERS = [
     "width-700|format-webp"
 ]
 
+Note that the ``srcSet`` attribute on ``ImageObjectType`` generates ``width-*`` filters, so if in use consider adding
+the relevant filters to the allowed list.
+
 
 DocumentObjectType
 ^^^^^^^^^^^^^^^^^^
--- a/docs/general-usage/graphql-types.rst
+++ b/docs/general-usage/graphql-types.rst
@@ -142,13 +142,29 @@ the ``ImageRenditionObjectType`` provides the following fields:
 ::
 
     id: ID
-    filterSpec: String!
+    url: String
     file: String!
     width: Int
     height: Int
-    focalPointKey: String!
+    aspectRatio: Float!
+    sizes: String!
     image: ImageObjectType!
-    url: String
+
+
+To prevent arbitrary renditions from being generated, set ``GRAPPLE_ALLOWED_IMAGE_FILTERS`` in your settings to a
+list of allowed filters. Read more about generating renditions in the Wagtail docs
+(`Generating renditions in Python <https://docs.wagtail.io/en/stable/advanced_topics/images/renditions.html#generating-renditions-in-python>` and
+`Using images in templates <https://docs.wagtail.io/en/stable/topics/images.html#image-tag>`
+
+For example:
+
+::
+
+GRAPPLE_ALLOWED_IMAGE_FILTERS = [
+    "width-1000",
+    "fill-300x150|jpegquality-60",
+    "width-700|format-webp"
+]
 
 
 DocumentObjectType
--- a/grapple/types/images.py
+++ b/grapple/types/images.py
@@ -62,8 +62,8 @@ def get_rendition_type():
 
 def rendition_allowed(rendition_filter):
     """Checks a given rendition filter is allowed"""
-    allowed_filters = getattr(settings, "GRAPPLE_ALLOWED_IMAGE_FILTERS", [])
-    if not allowed_filters:
+    allowed_filters = getattr(settings, "GRAPPLE_ALLOWED_IMAGE_FILTERS", None)
+    if allowed_filters is None or not isinstance(allowed_filters, list):
         return True
 
     return rendition_filter in allowed_filters
--- a/grapple/types/images.py
+++ b/grapple/types/images.py
@@ -1,5 +1,6 @@
 import graphene
 
+from django.conf import settings
 from graphene_django import DjangoObjectType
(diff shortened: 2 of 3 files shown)


zerolab:
@thibaudcolas can you take this for a spin?

ruisaraiva19 (approved):

docs/general-usage/graphql-types.rst
```diff
@@ -142,13 +142,29 @@ the ``ImageRenditionObjectType`` provides the following fields:
...
-    filterSpec: String!
+    url: String
     file: String!
     width: Int
     height: Int
-    focalPointKey: String!
+    aspectRatio: Float!
+    sizes: String!
     image: ImageObjectType!
-    url: String
```
thibaudcolas, line 151: question: I don’t think I understand those changes – is this just unrelated cleanup?

docs/general-usage/graphql-types.rst
```diff
@@ -142,13 +142,29 @@ the ``ImageRenditionObjectType`` provides the following fields:
...
+list of allowed filters. Read more about generating renditions in the Wagtail docs
+(`Generating renditions in Python <https://docs.wagtail.io/en/stable/advanced_topics/images/renditions.html#generating-renditions-in-python>` and
+`Using images in templates <https://docs.wagtail.io/en/stable/topics/images.html#image-tag>`
+
+For example:
+
+::
+
+GRAPPLE_ALLOWED_IMAGE_FILTERS = [
+    "width-1000",
```
thibaudcolas, line 164: question: Is it worth documenting that `srcSet` generates `width-*` filters, and as such also needs filtering? Or obvious enough?

thibaudcolas (commented):
Thank you Dan, all working pretty well for me. I was surprised by the behavior with `[]` as allowed_filters, will let you be the judge of whether this is worth addressing or not.

docs/general-usage/graphql-types.rst
zerolab: it is unrelated cleanup. noticed this while reading the docs. Was in the context and made sense to do it

docs/general-usage/graphql-types.rst
zerolab: that is a good suggestion

grapple/types/images.py
```diff
@@ -59,6 +60,15 @@ def get_rendition_type():
     return rendition_type
 
 
+def rendition_allowed(rendition_filter):
+    """Checks a given rendition filter is allowed"""
+    allowed_filters = getattr(settings, "GRAPPLE_ALLOWED_IMAGE_FILTERS", [])
+    if not allowed_filters:
```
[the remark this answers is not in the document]
zerolab: This is a great observation. In my experience I've yet to find a Wagtail project that does not make use of renditions, but from a logical point of view if one sets this to `[]` it would makes sense to interpret it as "no image filters are allowed"

docs/general-usage/graphql-types.rst
zerolab: Added a note about it

grapple/types/images.py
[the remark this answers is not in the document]
zerolab: changed a bit:
- any rendition is allowed if `GRAPPLE_ALLOWED_IMAGE_FILTERS` is not set, or if it is set but is not a list (probably overkill)
- then defer to checkign if filter is in allowed list. an empty list means no filters are allowed

grapple/types/images.py
```diff
@@ -59,6 +60,15 @@ def get_rendition_type():
     return rendition_type
 
 
+def rendition_allowed(rendition_filter):
+    """Checks a given rendition filter is allowed"""
+    allowed_filters = getattr(settings, "GRAPPLE_ALLOWED_IMAGE_FILTERS", None)
+    if allowed_filters is None or not isinstance(allowed_filters, list):
```
thibaudcolas, line 66:
```suggestion
    if allowed_filters is None or not isinstance(allowed_filters, (list, tuple)):
```

suggestion: it doesn’t cost much so we should also protect people from shooting themselves in the foot by using a tuple instead of the expected list (which would allow all filters with the current implementation).

thibaudcolas (commented):
Thank you @zerolab, all looking good to me, I’ve made one additional suggestion.

I’m also wondering whether the new "empty list" use case is worth a unit test – will leave it to you to make that call.

Otherwise all good to merge.

PR #165, opened 2021-03-10
Fix/empty list in structblock

bbliem:
When you have an empty list inside a StructBlock in a StreamField, it is resolved to None. I'm not completely sure if this is really a bug, but it looks a bit like it to me. I added a test for the presumed bug and a fix. I'm not sure if this might break something else, but at least the other tests succeed.


ruisaraiva19 (approved):

zerolab:
Thank you @bbliem, merged in 12f5977

PR #172, opened 2021-05-14
Support Wagtail v2.14

fabienheureux:
No changes seem to be required from Wagtail 2.13 release notes: https://docs.wagtail.io/en/stable/releases/2.13.html#updated-handling-of-non-required-streamfields

However, I noticed that `stream_data` was deprecated in 2.12, whereas 2.12 is a valid requirement for Grapple.
I blindly replaced `stream_data` by `raw_data` as suggested in the 2.12 release notes (https://docs.wagtail.io/en/stable/releases/2.12.html), however I would like to have a second pair of eyes here. 

Tests pass at the moment.


zerolab:
Thank you @fabienheureux

fwiw, `.stream_data` will be removed in 2.14 (https://github.com/wagtail/wagtail/blob/stable/2.12.x/wagtail/core/blocks/stream_block.py#L580)

I' would rather spend a bit more time and do this change properly  (as outlined in https://docs.wagtail.io/en/stable/releases/2.12.html#stream-data-on-streamfield-values-is-deprecated)

requirements.txt
```diff
@@ -1,5 +1,5 @@
 Django>=2.2,<3.2
-wagtail>=2.5,<2.12
+wagtail>=2.5,<=2.13
```
ruisaraiva19, line 2: Please follow the same convention as before:
```diff
-wagtail>=2.5,<=2.13
+wagtail>=2.5,<2.14
```

setup.cfg
```diff
@@ -37,7 +37,7 @@ include_package_data = true
 packages = find:
 install_requires =
     Django>=2.2, <3.2
-    wagtail>=2.5, <2.13
+    wagtail>=2.5, <=2.13
```
ruisaraiva19, line 40: Please follow the same convention as before:
```diff
-wagtail>=2.5, <=2.13
+wagtail>=2.5, <2.14
```

ruisaraiva19 (changes requested):

grapple/types/streamfield.py
```diff
@@ -22,7 +22,7 @@
 class GenericStreamFieldInterface(Scalar):
     @staticmethod
     def serialize(stream_value):
-        return stream_value.stream_data
+        return stream_value.raw_data
```
fabienheureux, line 25: @zerolab I might lack a bit of knowledge regarding StreamField representation. 
Can this remain as is or should I update as well ?

zerolab: `raw_data` is good.

zerolab: saying that, this will fail for anything < 2.12 (and we have a `wagtail>=2.5, <2.14` constraint).

We want to support 2.11+, at the very least, so a good old try/except should do the trick here

fabienheureux: Silly me 🤦 
Thanks for the review, I will update.

ruisaraiva19: @fabienheureux you can now merge the latest code from the `main` branch which contains an updated CI pipeline that runs tests on multiple Wagtail versions (2.9, 2.10, 2.11, 2.12).

After merging you can update the [wagtail versions list](https://github.com/GrappleGQL/wagtail-grapple/blob/main/.github/workflows/ci.yml#L32) to have version 2.13 as well!

ruisaraiva19: @fabienheureux let me know if you have any questions about the previous comment 😉

fabienheureux: Thanks for the suggestion, no question it was perfectly clear (just lacked a bit of free time to work on this PR lately).

fabienheureux: Test pass with all wagtail versions, I checked and it seems that `raw_data` property exists since a few Wagtail versions, for example here in 2.11 https://github.com/wagtail/wagtail/blob/1061caa5eff53f450520284b618e615fc2a809c8/wagtail/core/blocks/stream_block.py#L496

fabienheureux: After a bit of debugging, I don't get where this code is used. 
Any idea ?

fabienheureux:
I have a question about this conditionnal: https://github.com/GrappleGQL/wagtail-grapple/blob/main/grapple/types/streamfield.py#L111 

@zerolab do you have an idea of a query (in the example project) that could go through this conditionnal ? 
I am using this query 
```
{
  pages {
    ... on BlogPage {
      body {
        ...on StreamFieldInterface {
          rawValue
        }
      }
    }
  }
}
```

And can't go through. 
I am trying every streamfield block from the example project but I am not sure there is an existing block that matches this conditionnal.

I will continue digging this but maybe you have an idea.

fabienheureux:
I use this PR on an actively developped project (wagtail 2.13) and I did not encounter any issues yet. 
Would it be possible to have a second pair of eyes regarding my question above ? 

Does anyone have a project, using wagtail < 2.13 on which it could be tested ?

.github/workflows/ci.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       max-parallel: 4
       matrix:
         python-version: ['3.6', '3.7', '3.8', '3.9']
-        wagtail-version: ['2.9', '2.10', '2.11', '2.12']
+        wagtail-version: ['2.9', '2.10', '2.11', '2.12', '2.13']
```
zerolab, line 32:
```suggestion
        wagtail-version: ['2.11', '2.12', '2.13', '2.14']
```

grapple/types/streamfield.py
zerolab:
```suggestion
        try
            return stream_value.raw_data
        except AttributeError:
            return stream_value.stream_data
```

requirements.txt
```diff
@@ -1,5 +1,5 @@
 Django>=2.2,<3.2
-wagtail>=2.5,<2.12
+wagtail>=2.5,<2.14
```
zerolab, line 2:
```suggestion
wagtail>=2.11,<2.15
```

setup.cfg
```diff
@@ -37,7 +37,7 @@ include_package_data = true
 packages = find:
 install_requires =
     Django>=2.2, <3.2
-    wagtail>=2.5, <2.13
+    wagtail>=2.5, <2.14
```
zerolab, line 40:
```suggestion
    wagtail>=2.11, <2.15
```

zerolab (changes requested):
Hey @fabienheureux,

a couple of minor tweaks and we can get it in and do a release candidate

.github/workflows/ci.yml
fabienheureux: This will require a bit more changes because there are compability issues with `wagtail_grapple[channels]` since django 3.0 is required for wagtail 2.14 (https://docs.wagtail.io/en/latest/releases/2.14.html) 

But I think it is worth the change, thanks for the suggestion

fabienheureux:
> 
> 
> Hey @fabienheureux,
> 
> a couple of minor tweaks and we can get it in and do a release candidate

Thanks a lot for the suggestions !

ruisaraiva19 (approved):

PR #174, opened 2021-07-01
docs: schema mutation subscription hooks

fabienheureux:
Following @ruisaraiva19 on https://github.com/GrappleGQL/wagtail-grapple/pull/155, this PR add

- [x] docs
- [x] example for mutations
- [x] example for subscriptions
- [x] test for mutations
- [x] test for subscriptions


example/home/mutations.py
```diff
@@ -0,0 +1,31 @@
...
+        parent = graphene.Int()
+
+    ok = graphene.Boolean()
+    author = graphene.Field(
+        PageInterface,
+    )
+
+    def mutate(root, info, name, parent):
+        # We use uuid here in order to ensure the slug will always be unique accross tests
+        author = AuthorPage(name=name, title=name, slug=uuid.uuid4().hex[:6].upper())
```
zerolab, line 23: This probably warrants a note in the docs that any custom mutation should take care of validation

fabienheureux: In fact, validation errors are properly handled. 
I explicitely added this randomly generator slug to avoid validation errors, but an existing slug would result in the response below
```json
{
  "errors": [
    {
      "message": "{'slug': ['This slug is already in use']}",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "createAuthor"
      ]
    }
  ],
  "data": {
    "createAuthor": null
  }
}
```

example/requirements.txt
```diff
@@ -2,7 +2,7 @@
 black==19.3b0
 
 # Your app requirements.
--r ../requirements.txt
+-r ../requirements-channels.txt
```
fabienheureux, line 5: This was required to enable subscriptions in the example repo.

example/example/tests/test_grapple.py
```diff
@@ -59,9 +60,11 @@ def test_pages(self):
 
         executed = self.client.execute(query)
 
-        self.assertEquals(type(executed["data"]), dict_type)
+        self.assertTrue(isinstance(executed["data"], collections.OrderedDict))
```
fabienheureux, line 63: This started failing after using `requirements-channels.txt` file for dependencies with 
```
    self.assertEquals(type(executed["data"]), dict_type)
AssertionError: <class 'collections.OrderedDict'> != <class 'dict'>
```

Let me know if this need to be addressed in another PR

fabienheureux:
@ruisaraiva19 @zerolab I added tests for subscriptions, an example in the example folder and in the docs. 
This is ready for review. 

One thing: some tests started failing after updating the dependencies (I think, but I am not 100% sure they started failing after this event). I would be happy to have your thoughts on these failing tests...(comments in "Files changed" tab ;) )

example/example/tests/test_grapple.py
zerolab: That has to do with graphql-core versions, I believe. 

https://github.com/GrappleGQL/wagtail-grapple/blob/cdc0d1d94a5ac7107e8563929ea185db5bce3787/example/example/tests/test_grapple.py#L9-L12, via https://github.com/GrappleGQL/wagtail-grapple/pull/105/commits/2313c52c41796075c26ac6f1159b4f1f75fd2ef7

example/home/test/test_general.py
```diff
@@ -213,4 +216,30 @@ def test_mutation(self):
         # Now we ensure that AuthorPage-specific fields are well returned
         self.assertIn("name", data["author"])
 
+        # Finally, we ensure that data passed in the first place to the query are indeed
+        # returned after the author has been saved to the database.
         self.assertEqual(data["author"]["name"], self.name)
+        self.assertEqual(data["author"]["slug"], self.slug)
+
+
+class TestRegisterSubscription(BaseGrappleTest):
```
zerolab, line 226: because the channels implementation has different requirements (Django < 3.x), I would say move this to its own test and only call in https://github.com/GrappleGQL/wagtail-grapple/blob/main/.github/workflows/ci.yml#L52-L65, which annoyingly has a missing step 🤦🏼  (https://github.com/GrappleGQL/wagtail-grapple/blob/main/.github/workflows/ci.yml#L45-L50)

example/example/tests/test_grapple.py
fabienheureux: It does, I reverted this change

example/home/test/test_general.py
fabienheureux: https://github.com/GrappleGQL/wagtail-grapple/pull/174/files#diff-1db27d93186e46d3b441ece35801b244db8ee144ff1405ca27a163bfe878957fR51
I added the missing step here 

https://github.com/GrappleGQL/wagtail-grapple/pull/174/files#diff-b2d9000ea2939f79f0b93f509f0fcbf9cf7d88604d40b8ee1705a699bbe3040e
And a dedicated test for subscriptions here

fabienheureux:
I have updated the tests, I am not very proud of the way I don't run the test when channels is not enabled, if you have a better suggestion I would be happy to change this.

.github/workflows/tests.yml
```diff
@@ -47,3 +47,10 @@ jobs:
         run: |
           pip install -r requirements-channels.txt
           pip install .
+
+      - name: Run Tests
```
ruisaraiva19, line 51: @zerolab we didn't run tests before with channels enabled before. Any reason for that?
If so, this would be removed.

ruisaraiva19 (approved):

ruisaraiva19:
@fabienheureux can you fix the merge conflicts?

.github/workflows/tests.yml
zerolab: it was a glaring omission 🤦‍♂️

ruisaraiva19: ok!

docs/general-usage/hooks.rst
```diff
@@ -31,4 +35,67 @@ replace or remove any of the default mixins that are not of use in your project:
...
+
+    @hooks.register("register_schema_mutation")
+    def register_author_mutation(mutation_mixins):
+        mutation_mixins.append(Mutation)
+
+
+Subscription
+^^^^^^^^^^^^
+
+Grapple provides a ``register_schema_subscription`` hook that is called when it creates the schema. You can use it to add your custom ``Subscription`` mixins
```
zerolab, line 81: Worth making a note that this is enabled only when installed with `wagtail_grapple[channels]`

fabienheureux:
I have fixed the conflicts and added a note regarding subscriptions being enabled only when Grapple is installed with Django Channels

ruisaraiva19 (approved):

zerolab (approved):

PR #175, opened 2021-07-04
feat: single settings GRAPPLE support

ruisaraiva19:
- [x] Grapple setting object
    - [x] backwards compatibility
    - [x] deprecated settings warnings
    - [x] removed settings errors
- [x] documentation
    - [x] all settings available
    - [x] settings defaults
- [x] tests

All Wagtail Grapple settings are now done via the `GRAPPLE` settings object.

```python
# Your project settings file
GRAPPLE = {
    "APPS": ["home", "images", "documents"],
    "AUTO_CAMELCASE": True,
    "ADD_SEARCH_HIT": False,
    "EXPOSE_GRAPHIQL": False,
    "ALLOWED_IMAGE_FILTERS": [
        "width-1000",
        "fill-300x150|jpegquality-60",
        "width-700|format-webp",
    ],
}
```

This adds support for customizing the pagination `limit`/`per_page` default and maximum values.
This can be done by using the following settings:
```python
# Your project settings file
GRAPPLE = {
    # ...
    "PAGE_SIZE": 15, # default 10
    "MAX_PAGE_SIZE": 30, # default 100
}
```

### DIFF
--- a/docs/general-usage/decorators.rst
+++ b/docs/general-usage/decorators.rst
@@ -174,6 +174,10 @@ You can now query your adverts with the following query:
         }
     }
 
+The default ``per_page`` value is 10 and can be changed with the ``GRAPPLE_PAGE_SIZE`` setting.
+
+The ``per_page`` has a maximum value of 100 by default and can be changed with the ``GRAPPLE_PAGE_SIZE_MAX`` setting.
+
 You can add custom query parameters like so:
 
 ::
--- a/docs/general-usage/decorators.rst
+++ b/docs/general-usage/decorators.rst
@@ -174,9 +174,16 @@ You can now query your adverts with the following query:
         }
     }
 
-The default ``per_page`` value is 10 and can be changed with the ``GRAPPLE_PAGE_SIZE`` setting.
+The default ``per_page`` value is 10 and can be changed with the ``GRAPPLE["PAGE_SIZE"]`` setting.
+The ``per_page`` has a maximum value of 100 by default and can be changed with the ``GRAPPLE["MAX_PAGE_SIZE"]`` setting.
 
-The ``per_page`` has a maximum value of 100 by default and can be changed with the ``GRAPPLE_PAGE_SIZE_MAX`` setting.
+::
+
+    GRAPPLE = {
+        # ...
+        "PAGE_SIZE": 10,
+        "MAX_PAGE_SIZE": 100,
+    }
 
 You can add custom query parameters like so:
 
--- a/docs/general-usage/settings.rst
+++ b/docs/general-usage/settings.rst
@@ -1,0 +1,124 @@
+Settings
+========
+
+Configuration for Wagtail Grapple is all namespaced inside a single Django setting,
+named ``GRAPPLE``.
+
+For example your project's ``settings.py`` file might include something like this:
+
+::
+
+    GRAPPLE = {
+        'APPS': ['home'],
+        'ADD_SEARCH_HIT': True,
+    }
+
+
+Accessing settings
+------------------
+
+If you need to access the values of Wagtail Grapple's settings in your project, you should use the
+``grapple_settings`` object. For example.
+
+::
+
+    from grapple.settings import grapple_settings
+
+    print(grapple_settings.APPS)
+
+The ``grapple_settings`` object will check for any user-defined settings, and otherwise fall back to
+the default values.
+
+
+API Reference
+-------------
+
+
+Grapple settings
+^^^^^^^^^^^^^^^^
+
+APPS
+****
+
+A list/tuple of the apps that Grapple will scan for models to generate GraphQL types that adopt their structure.
+
+Default: ``['home']``
+
+
+AUTO_CAMELCASE
+**************
+
+By default, all field and argument names will be converted from `snake_case` to `camelCase`.
+To disable this behavior, set the ``GRAPPLE['AUTO_CAMELCASE']`` setting to `False`.
+
+Default: ``True``
+
+
+EXPOSE_GRAPHIQL
+***************
+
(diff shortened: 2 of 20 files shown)


example/example/settings/base.py
```diff
@@ -173,6 +173,8 @@
 GRAPHENE = {"SCHEMA": "grapple.schema.schema"}
 GRAPPLE_APPS = {"images": "", "home": "", "documents": ""}
 GRAPPLE_ADD_SEARCH_HIT = True
+GRAPPLE_PAGE_SIZE = 10
+GRAPPLE_PAGE_SIZE_MAX = 50
```
zerolab, line 177: suggestion: worth lowering these both, say 5 and 6 (instead of 10 and 50) so the max page size can be tested

docs/general-usage/decorators.rst
```diff
@@ -174,6 +174,10 @@ You can now query your adverts with the following query:
         }
     }
 
+The default ``per_page`` value is 10 and can be changed with the ``GRAPPLE_PAGE_SIZE`` setting.
```
zerolab, line 177: suggestion: looks like it may be time to move to a single settings approach. Django Rest Framework and others do it. Wagtail itself will eventually move to that as well.

With these we now have (from the example settings):

```python
GRAPPLE_APPS = {"images": "", "home": "", "documents": ""}
GRAPPLE_ADD_SEARCH_HIT = True
GRAPPLE_PAGE_SIZE = 10
GRAPPLE_PAGE_SIZE_MAX = 50
```

Would look so much better as:
```python
GRAPPLE = {
    "apps": {"images": "", "home": "", "documents": ""},
    "add_search_hit": True,
    "page_size": 10,
    "page_size_max": 50,   # `max_page_size` ?
}
```

would be amazing if we could support a list for Grapple apps (i.e. `["images", "home", "documents"]`) as well -- the current dict with empty values is a long bugbear of mine -- but not imperative on this PR

zerolab (changes requested):
Hey @ruisaraiva19,

Thank you for these. They are sane changes. We do, however start moving into the "lots of settings" land and would be good to rationalise them.

docs/general-usage/decorators.rst
ruisaraiva19: Makes sense, I'll update the code accordingly.

example/example/settings/base.py
ruisaraiva19: Yeah, I'll add some tests later today.

docs/general-usage/decorators.rst
zerolab: 🌟

ruisaraiva19: @zerolab updated to use single object settings approach!

ruisaraiva19: I just added a `settings` module to easily manage all Grapple settings. It's inspired by Django Rest Framework's settings module. 👀

docs/general-usage/settings.rst
```diff
@@ -0,0 +1,124 @@
...
+
+Grapple settings
+^^^^^^^^^^^^^^^^
+
+APPS
+****
+
+A list/tuple of the apps that Grapple will scan for models to generate GraphQL types that adopt their structure.
+
+Default: ``['home']``
```
zerolab, line 45: This default makes a lot of assumptions. I would keep it empty

grapple/settings.py
```diff
@@ -0,0 +1,84 @@
...
+This module provides the `grapple_settings` object, that is used to access
+Wagtail Grapple settings, checking for user settings first, then falling
+back to the defaults.
+"""
+from django.conf import settings
+from django.test.signals import setting_changed
+
+
+DEFAULTS = {
+    "APPS": ["home"],
```
zerolab, line 17: as above, this default makes a lot of assumptions. I would keep it empty

grapple/settings.py
```diff
@@ -0,0 +1,84 @@
...
+        return self._user_settings
+
+    def __getattr__(self, attr):
+        if attr not in self.defaults:
+            raise AttributeError("Invalid Grapple setting: '%s'" % attr)
+
+        try:
+            # Check if present in user settings
+            val = self.user_settings[attr]
+        except KeyError:
```
zerolab, line 58: This is an excellent solution for the single setting approach!

We have a "beta"  pypi classifier (and by semver standards not being a full major version) which in a way gives us freedom to introduce breaking changes, I feel that the package is used by enough production sites that we should follow a 2 release deprecation process.

As such, would be good to try to support the old settings and we will make a note in the release notes about them being deprecated and pending removal in the following release.

so I would add a bit more code here that tries to get the value from the separate settings

docs/getting-started/installation.rst
```diff
@@ -10,57 +10,53 @@ settings file:
...
+    urlpatterns = [
+        ...
+        url(r"", include(grapple_urls)),
+        ...
+    ]
 
 Done! Now you can proceed onto configuring your models to generate
 GraphQL types that adopt their structure.
 
-By default, all field and argument names will be converted from `snake_case`
```
zerolab, line 60: suggestion: link out to the settings doc page for the full set.

zerolab (changes requested):
@ruisaraiva19 this PR is full of great stuff!

A few changes needed: a couple of minor ones, plus fallback support for the old-style settings as I feel enough builds use this in production that we should follow a 2 release deprecation procedure

grapple/settings.py
ruisaraiva19: ok, I'll add that logic.

ruisaraiva19: @zerolab I just added `deprecated` and `removed` settings support. Deprecated settings will log a warning and removed settings will raise a `RuntimeError`.

zerolab (approved):
Stellar! :shipit:

PR #176, opened 2021-07-07
ci: run tests on multiple wagtail versions

ruisaraiva19:
This runs tests on Wagtail version `2.9`, `2.10`, `2.11`, and `2.12`.

### DIFF
--- a/example/home/migrations/0002_create_homepage.py
+++ b/example/home/migrations/0002_create_homepage.py
@@ -1,9 +1,6 @@
 # -*- coding: utf-8 -*-
 from django.db import migrations
-import wagtail
-
-
-wagtail_version = tuple(map(int, wagtail.__version__.split(".")))
+from wagtail import VERSION as WAGTAIL_VERSION
 
 
 def create_homepage(apps, schema_editor):
@@ -56,7 +53,7 @@ class Migration(migrations.Migration):
         [
             ("wagtailcore", "0053_locale_model"),
         ]
-        if wagtail_version >= (2, 11)
+        if WAGTAIL_VERSION >= (2, 11)
         else []
     )
 
--- a/example/home/migrations/0002_create_homepage.py
+++ b/example/home/migrations/0002_create_homepage.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 from django.db import migrations
+import wagtail
 
 
 def create_homepage(apps, schema_editor):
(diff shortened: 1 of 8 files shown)


zerolab:
Looks like the tests will need updating:
there's the migration dependency on `0053_locale_model` which came in with 2.11
then there is the `get_document_model_string` import error and I suspect a couple of others relating to deprecations

example/home/migrations/0002_create_homepage.py
```diff
@@ -53,7 +56,7 @@ class Migration(migrations.Migration):
         [
             ("wagtailcore", "0053_locale_model"),
         ]
-        if wagtail.__version__ >= "2.11"
+        if wagtail_version >= (2, 11)
```
zerolab, line 59: you could import https://github.com/wagtail/wagtail/blob/main/wagtail/__init__.py#L10 (`from wagtail import VERSION as WAGTAIL_VERSION`) and then `if WAGTAIL_VERSION >= (2, 11)

example/home/test/test_blog.py
```diff
@@ -426,8 +427,12 @@ def test_blog_embed(self):
             if block["blockType"] == "VideoBlock":
                 embed = block["youtubeLink"]
                 self.assertTrue(isinstance(embed["url"], str))
-                self.assertEquals(embed["embed"], raw_embed["html"])
-                self.assertEquals(embed["rawEmbed"], json.dumps(raw_embed))
+                if WAGTAIL_VERSION >= (2, 11):
+                    self.assertEquals(embed["embed"], raw_embed["html"])
+                    self.assertEquals(embed["rawEmbed"], json.dumps(raw_embed))
+                else:
+                    self.assertEquals(embed["embed"], None)
```
zerolab, line 434: question: why is this the case for <= 2.11?

could probably use a comment. future us will thank you ;)

zerolab (approved):
Other than that one final question, this LGTM

Thank you @ruisaraiva19 !

example/home/test/test_blog.py
ruisaraiva19: @zerolab I just remembered [this issue](https://github.com/wagtail/wagtail/issues/6656#issuecomment-749711713) and now I [added](https://github.com/GrappleGQL/wagtail-grapple/pull/176/commits/ff78bb127e6a30c42eb8effd6681bae0e04024d5) a custom `WAGTAILEMBEDS_FINDERS` setting when running with Wagtail version < 2.11 and now the tests pass as expected without the if statement.

PR #177, opened 2021-07-08
feat: add middleware support for queries

ruisaraiva19:
This adds middleware support for query fields.
- [x] query field middleware
    - [x] register_query_field
    - [x] register_paginated_query_field
    - [x] register_singular_query_field
- [x] documentation
- [x] tests


docs/general-usage/middleware.rst
```diff
@@ -0,0 +1,71 @@
...
+Resolve arguments
+-----------------
+
+Middlewares ``resolve`` is invoked with several arguments:
+
+- ``next`` represents the execution chain. Call ``next`` to continue evaluation.
+- ``root`` is the root value object passed throughout the query.
+- ``info`` is the resolver info.
+- ``args`` is the dict of arguments passed to the field.
+
```
zerolab, line 23: This is a copy of https://docs.graphene-python.org/en/latest/execution/middleware/.
Maybe condense it to the first part and link out to the Graphene docs?

docs/general-usage/middleware.rst
```diff
@@ -0,0 +1,71 @@
...
+
+GrappleMiddleware
+-----------------
+.. module:: grapple.middleware
+.. class:: GrappleMiddleware(object)
+
+::
```
zerolab, line 30: nitpick: I prefer the `.. code-block:: python` notation as we get nicer syntax highlighting

note that I didn't raise it for the settings docs as they were small enough, but I think little details like this make for a nicer reading experience.

Also it is worth adding a `# settings.py` line before `GRAPHENE = {` to softly indicate this goes into settings, unless you spell it out

grapple/actions.py
```diff
@@ -23,7 +23,7 @@
 from .types.images import ImageObjectType
 from .types.pages import PageInterface, Page
 from .types.streamfield import generate_streamfield_union
-from .helpers import streamfield_types
+from .helpers import streamfield_types, fields_middlewares
```
zerolab, line 26:
```suggestion
from .helpers import streamfield_types, field_middlewares
```

it is a nitpick. `field_middlwares` reads easier than `fields_middlewares`

zerolab (commented):
A few minor suggestions. Quality work!
@fabienheureux do you have any time to give it a quick scan?

example/home/test/test_general.py
```diff
@@ -18,47 +35,62 @@ def query():
...
+            firstPost {
+                id
+            }
+        }
+        """
+        self.request.user = AuthenticatedUser()
+        results = self.client.execute(query, context_value=self.request)
+
+        data = results["data"]["firstPost"]
+        self.assertEqual(data, None)
```
fabienheureux, line 67: **Question :** do we really want the data to be `None` here ?

example/home/test/test_general.py
```diff
@@ -71,53 +103,95 @@ def test_query_field_plural(self):
...
+        {
+            posts {
+                id
+            }
+        }
+        """
+        self.request.user = AuthenticatedUser()
+        results = self.client.execute(query, context_value=self.request)
+        data = results["data"]["posts"]
+        self.assertEqual(data, None)
```
fabienheureux, line 124: ditto

fabienheureux:
A few questions regarding tests, thanks for this PR @ruisaraiva19 this is a very good improvement for Grapple !! 🎉

example/home/test/test_general.py
ruisaraiva19: @fabienheureux yes, because we have the [`IsAnonymousMiddleware`](https://github.com/GrappleGQL/wagtail-grapple/pull/177/files/cca22ad77fed9ea578e9e7d2567471b55b776f56#diff-d3f3dfeadc002df554ce4060d35a5ec429f66733c519d726ba80e447bc2e3062R72) being used on this query. So if we are authenticated, the query will return `None`.

example/home/test/test_general.py
ruisaraiva19: The same applies here regarding the `IsAnonymousMiddleware`.

zerolab (approved):

example/home/test/test_general.py
fabienheureux: Ah. Makes sense. 
I assumed `IsAnonymousMiddleware` would also allow `Authenticated` users, but it is worthless as default queries are available to every users 😅

fabienheureux (approved):

PR #180, opened 2021-08-08
add support for specifying custom resolvers

### LINKED ISSUE #179: Add ability to use custom resolvers in the various GraphQLFields
Oftentimes, one needs to preprocess some data before passing it to graphql, or just implement a fully cusomized property. For django models, this is possible by adding a `@property` and pointing a `GraphQLField` to it. However, for, say, custom `StructBlocks` (that were registered using `@register_streamfield_block`) there is currently no way to do that. More concretely, I would like to render the HTML output of Wagtail-markdown in a StructBlock to a GraphQL field.

One way to do that would be to add a possibility to pass a custom resolver to the GraphQLField, which is then taken instead of the default `streamfield_resolver` in actions.py, line 395. I could submit a PR if wanted.

ARogovskyy:
Adds basic support for custom resolvers. It would be nice if specifying custom arguments was supported too, but for that we probably need to rethink the design (how/where to specify arguments etc.)
closes #179

### DIFF
--- a/grapple/actions.py
+++ b/grapple/actions.py
@@ -166,8 +166,14 @@ def get_field_type(field):
         else:
             return field, graphene.Field(field_type)
 
+def get_wrapped_custom_resolver(field):
+    if not field.resolver:
+        return None
+    return lambda self, instance, info, **kwargs: field.resolver(instance, info, **kwargs)
 
 def model_resolver(field):
+    if field.resolver:
+        return get_wrapped_custom_resolver(field)
     def mixin(self, instance, info, **kwargs):
         from .utils import resolve_queryset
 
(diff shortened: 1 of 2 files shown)


zerolab:
@ARogovskyy we have [pre-commit](https://pre-commit.com/) to help for linting, I suggest using that:

```sh
# install pre-commit: usually `pip install pre-commit`
# then in repository root:
$ pre-commint install
$ pre-commit run --all-files
```

zerolab:
@ARogovskyy in addition to the linter, it would be great to have at least one test and an update to the documentation.
Thank you!

zerolab:
Closing in favour of #188

PR #182, opened 2021-08-18
Low quality image placeholder for images

fabienheureux:
## Description 

Add `placeholder_blur` (low quality image placeholder) field to images, so that a blurred rendition of image can be generated on the fly.
A base64 version of the image is returned, to use for example with next/js blurDataURL props https://nextjs.org/docs/api-reference/next/image#blurdataurl

## TODO
- [x] Add tests
- [x] Add docs
- [ ] Improve wagtail 2.15 "support"
- [ ] Prevent the generation of blurred image to be CPU-consuming
- [ ] Find the best image size to use
- [ ] Test GIF support


fabienheureux:
@zerolab @ruisaraiva19 before I spend more time on this, do you think this feature could be in the scope of Grapple ? 
If so, I will add docs/tests (and put it back to draft).

example/images/models.py
```diff
@@ -33,4 +33,5 @@ def custom_rendition_property(self, info, **kwargs):
         GraphQLString("height", required=True),
         GraphQLImage("image", required=True),
         GraphQLString("file", required=True),
+        GraphQLString("lqip", required=True),
```
zerolab, line 36: note: I am not a fan of using `lqip` here. too much cognitive overhead. maybe `blur_hash` or `blur_placeholder`?

grapple/types/images.py
```diff
@@ -175,3 +197,23 @@ def resolve_image_type(self, info, **kwargs):
...
+    # Wand modifies images in place so clone it first to prevent
+    # altering the original image
+    blurred_image = image.image.clone()
+    blurred_image.gaussian_blur()
+    return WandImage(blurred_image)
+
+
+# Register the operations in Willow
+willow_registry.register_operation(PillowImage, "blur", pillow_blur)
+willow_registry.register_operation(WandImage, "blur", wand_blur)
```
zerolab, line 219: suggestion: namspace this to something like `grapple_blur` in the event Willow adds its own blur operation, or site implementer for that matter

grapple/types/images.py
```diff
@@ -23,6 +28,23 @@ class BaseImageObjectType(graphene.ObjectType):
...
     collection = graphene.Field(lambda: CollectionObjectType, required=True)
+    lqip = graphene.String()
+
+    def resolve_lqip(self, info, **kwargs):
+        """
+        Get a low quality image placeholder of the original image.
+        This can directly be use in frontend frameworks like next.js
+        """
+        rendition = self.get_rendition("width-64|jpegquality-50")
+        with rendition.get_willow_image() as willow:
```
zerolab, line 39: this could be done without the Wagtail PR via a convenience method

example/images/models.py
fabienheureux: blur_hash is great, but it might lead to confusion because of https://github.com/woltapp/blurhash 
Do you think this could b an issue ?

zerolab: heh, it could. `placeholder_blur` then?

fabienheureux: placeholder_blur it is then 👍

docs/general-usage/usage-with-frontend-frameworks.rst
```diff
@@ -0,0 +1,26 @@
+Usage with frontend frameworks
+##################################
```
zerolab, line 2:
```suggestion
##############################
```

docs/general-usage/usage-with-frontend-frameworks.rst
```diff
@@ -0,0 +1,26 @@
...
+
+A few features of Wagtail Grapple are useful for specific frontend frameworks.
```
zerolab, line 4:
```suggestion
Several Wagtail Grapple are useful for specific frontend frameworks.
```

grapple/types/images.py
```diff
@@ -22,7 +27,25 @@ class BaseImageObjectType(graphene.ObjectType):
...
+
     collection = graphene.Field(lambda: CollectionObjectType, required=True)
+    placeholder_blur = graphene.String()
+
+    def resolve_placeholder_blur(self, info, **kwargs):
+        """
+        Get a low quality image placeholder of the original image.
+        This can directly be use in frontend frameworks like next.js
+        """
+        rendition = self.get_rendition("width-64|jpegquality-50")
```
zerolab, line 39: question: could the quality go lower?

zerolab (commented):
A few suggestions.

I see the sqlite db has changes... maybe drop them since there are no new migrations

grapple/types/images.py
fabienheureux: It could, we could even use smaller image in size. 
I did not use this behaviour a lot, it might requires some tweaking in the long run. 

I can provide a few examples first to compare various sizes/qualities. Let me know what you prefer

grapple/types/images.py
```diff
@@ -22,7 +27,25 @@ class BaseImageObjectType(graphene.ObjectType):
...
+        with get_willow_image(rendition) as willow:
+            i = willow.grapple_blur()
+            tf = tempfile.SpooledTemporaryFile()
+            i.save_as_png(tf)
+            tf.seek(0)
+            hash = base64.b64encode(tf.read())
+            tf.close()
+
+        return "data:image/jpeg;base64,%s" % hash.decode("utf-8")
```
zerolab, line 48: on L43 we have `i.save_as_png(tf)`, but this says is it `image/jpeg` 🤔

zerolab: @fabienheureux I'd like to get this in before making a new release. any chance you could have a look at this?

and possibly add a comment to `SourceImageIOError`, `get_willow_image` and `open_file` that we should remove then when we drop support for Wagtail < 2.15

zerolab: actually, having thought more of it - this will create a temp file and blur it every single time which means this is potentially expensive on image heavy site 🤔

One way around it would be to register a new image filter so that the whole thing is saved as a rendition so we only do the base64 encoding

fabienheureux: Hey @zerolab, I put this PR back into draft mode, there are a few areas of improvements, especially finding the best rendition's size, and as you suggested, create a new image filter. 

I use this PR on production on a not-so-image-heavy project, and I noticed CPU usage peaks.
I did not have time to investigate yet but it might be related.

fabienheureux: Quick update on this PR, apart from the performance issues, it can break some features if the website uses GIF images. 
I am keeping this PR on hold before I find the time to test it properly, as it has many side effect I might have not think of.

fabienheureux:
@zerolab I don't think I will find the time to finish this PR in the following weeks. 
This feature was abandonned by the client I was developping it for. 
 
Should we close it or keep it as a draft until someone (maybe me later) picks it up ?

PR #184, opened 2021-08-20
Fix versions solving issues introduced by wagtail 2.14 support

fabienheureux:
# Fix versions solving issues introduced by wagtail 2.14 support
There are currently compatibility issues between Wagtail 2.14 (django > 3.0) and channels 1 and its dependencies.
This PR attempts to solve these issues. 

## `graphql-ws`
The package has not been published but has some useful changes on the master branch.
Using it can help simplify a lot the configuration of subscriptions for Grapple, cf https://github.com/graphql-python/graphql-ws

## Routing
These changes use the default `graphql-ws` routing but we could use a custom one  
https://github.com/graphql-python/graphql-ws/blob/master/graphql_ws/django/routing.py

## Channels support
`channels` latest version is the version 3, but `graphql-ws` is not fully compatible with `channels>2`. 
This PR upgrades channels and graphql-ws (using master branch), and update its usage in Grapple.

I opened a PR on `graphql-ws` repo to bring `channels>=3` support: https://github.com/graphql-python/graphql-ws/pull/76

Fixes https://github.com/GrappleGQL/wagtail-grapple/issues/86

## Potential breaking change
This drops channels 1 support as it would be useful only if we supported python < 3
Let me know if we need to keep compatibility with channels <2, I might not have the big picture here. 

Note: it could be worth using poetry or a python package manager to prevent these versions solving issues in the future (Poetry detected the issue in my case).


requirements-channels.txt
```diff
@@ -1,14 +1,6 @@
...
-wagtail-headless-preview==0.1.4
+-r ./requirements.txt
 
 # Channels-specific requirements
-channels==1.1.8
-asgi_redis
-graphql_ws
+channels==3.0.4
+channels_redis==3.3.0
+git+git://github.com/fabienheureux/graphql-ws@fix/channels-3-compatibility#egg=graphql-ws
```
fabienheureux, line 6: TODO: Remove before merge

grapple/urls.py
```diff
@@ -23,7 +15,7 @@ def graphiql(request):
         "supports_subscriptions": has_channels,
     }
     if has_channels:
-        graphiql_settings["SUBSCRIPTIONS_TRANSPORT_VERSION"] = "0.9.19"
+        graphiql_settings["SUBSCRIPTIONS_TRANSPORT_VERSION"] = "0.8.3"
```
fabienheureux, line 18: I had to downgrade the version of this script because it lead to errors in graphiql

zerolab: worth adding a comment/todo there to that effect

setup.cfg
```diff
@@ -45,7 +45,7 @@ install_requires =
 
 [options.extras_require]
 channels =
-    Django>=2.2, <2.3
-    channels==1.1.8
-    asgi_redis
+    Django>=2.2, <3.2
```
zerolab, line 48: why not 3.2?

fabienheureux: Good question, I might have push this change this by mistake while trying to solve the version conflict. 

In requirements.txt there is a restriction on django version though (< 3.2), should we remove the restriction there as well ?

zerolab: we absolutely should.

zerolab:
>  Let me know if we need to keep compatibility with channels <2, I might not have the big picture here.
I think we're ok to drop it especially that Wagtail 2.15 is not that far off and with its release we'll be dropping support for 2.11 as well.
People can still use the current release with Django < 3.0

fabienheureux:
I am now an admin on graphql-ws and my PR has been merged https://github.com/graphql-python/graphql-ws/commit/474f733761c04f39ccd9367190b5a85a5033fc2c 
I will release a new version of graphql-ws on PyPi later this week, it might be worth waiting for this release to be published before merging the current M in order to not rely on the commit hash in the requirements file

fabienheureux:
I published a new `graphql-ws` version on PyPi a few minutes ago. 
Tests now pass, I think we can merge this PR @ruisaraiva19 @zerolab

setup.cfg
```diff
@@ -45,7 +45,7 @@ install_requires =
 
 [options.extras_require]
 channels =
-    Django>=2.2, <2.3
-    channels==1.1.8
-    asgi_redis
-    graphql_ws
+    Django>=2.2
+    channels==3.0.4
```
zerolab, line 49: https://github.com/GrappleGQL/wagtail-grapple/pull/184/files#diff-e76e3d4b8b9857a8082a63501a2453c42fc867bd2900f1cc814c2b87c4feb9ccR5 says `channels==2.4.0` which is different from this 🤔

example/example/settings/base.py
```diff
@@ -212,18 +212,12 @@
...
-            "BACKEND": "asgiref.inmemory.ChannelLayer",
-            "ROUTING": "grapple.urls.channel_routing",
-        }
-    }
+    import channels
+
+    INSTALLED_APPS += [
+        "channels",
+    ]
+    ASGI_APPLICATION = "graphql_ws.django.routing.application"
```
zerolab, line 220: The README/docs need updating to reflect this too

fabienheureux: I updated these in the preview section as it is at the moment
- https://wagtail-grapple.readthedocs.io/en/latest/getting-started/installation.html#installation 
- https://wagtail-grapple.readthedocs.io/en/latest/general-usage/preview.html

Should I add a section in the installation section ?
If you think I should, it might be better to have a dedicated section about "How to enable channels" because it might be a bit confusing to mix default installation steps with channel-enabled steps.

zerolab: a separate sections sounds great

fabienheureux: Thanks for your feedback, I will add it tomorrow.

setup.cfg
```diff
@@ -46,6 +46,6 @@ install_requires =
 [options.extras_require]
 channels =
     Django>=2.2
-    channels==3.0.4
+    channels==2.4.0
```
zerolab, line 49: Looking at https://github.com/django/channels/blob/main/CHANGELOG.txt maybe it should stay `>=3.0, <3.1`?

setup.cfg
```diff
@@ -45,7 +45,7 @@ install_requires =
 
 [options.extras_require]
 channels =
-    Django>=2.2, <2.3
-    channels==1.1.8
-    asgi_redis
-    graphql_ws
+    Django>=2.2
+    channels>=3.0<3.1
```
zerolab, line 49:
```suggestion
    channels>=3.0, <3.1
```

zerolab (changes requested):
One final thing to fix (the channels versions in setup.cfg) and this is good to go!

fabienheureux:
Thanks for taking the time to provide a suggestion @zerolab, this week was quite busy 😅.

zerolab (approved):

fabienheureux:
@zerolab @ruisaraiva19 is there anything needed on my side to get this PR merged ? 
Thanks !

zerolab:
@fabienheureux there was nothing preventing it. Merged. Thank you for the nudge and your hard work on this! 🎉

isolationism:
Thank you for getting to this as I haven't been able to. Wondered about the backward compatibility to older versions myself, seems like that isn't an issue moving forward. Awesome!

PR #186, opened 2021-09-15
Update middleware doc

fabienheureux:
`graphene-django` expect this setting to be a list and not a string.
https://github.com/graphql-python/graphene-django/blob/623d0f219ebeaf2b11de4d7f79d84da8508197c8/graphene_django/views.py#L50

Using a single string raised an exception.

### DIFF
--- a/docs/general-usage/middleware.rst
+++ b/docs/general-usage/middleware.rst
@@ -18,7 +18,7 @@ GrappleMiddleware
     # settings.py
     GRAPHENE = {
         # ...
-        'MIDDLEWARE': ['grapple.middleware.GrappleMiddleware'],
+        "MIDDLEWARE": ["grapple.middleware.GrappleMiddleware"],
     }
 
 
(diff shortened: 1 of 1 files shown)


docs/general-usage/middleware.rst
```diff
@@ -18,7 +18,7 @@ GrappleMiddleware
     # settings.py
     GRAPHENE = {
         # ...
-        'MIDDLEWARE': 'grapple.middleware.GrappleMiddleware',
+        'MIDDLEWARE': ['grapple.middleware.GrappleMiddleware'],
```
zerolab, line 21: Since we're using black for formatting

```suggestion
        "MIDDLEWARE": ["grapple.middleware.GrappleMiddleware"],
```

other than that, LGTM

fabienheureux: I lazily edited from github directly...I will avoid this in the future 🙄 
Thanks

PR #194, opened 2021-11-01
fix(graphiql): use correct endpoint url for GraphiQL

### LINKED ISSUE #192: graphiql endpointURL should be configurable
In https://github.com/GrappleGQL/wagtail-grapple/blob/main/grapple/urls.py#L14 the graphql-endpoint it set statically to `/graphql`. The default configuration used in doc is exposing `/api/graphql`. This is a mismatch and cannot work together.

Maybe pointing the graphiql endpoint to `grapple_urls.root `is more conform?

ruisaraiva19:
- Closes #192

### DIFF
--- a/grapple/urls.py
+++ b/grapple/urls.py
@@ -1,4 +1,5 @@
 from django.shortcuts import render
+from django.urls import reverse
 from django.conf.urls import url
 from django.views.decorators.csrf import csrf_exempt
 
@@ -11,7 +12,7 @@ def graphiql(request):
     graphiql_settings = {
         "REACT_VERSION": "16.14.0",
         "GRAPHIQL_VERSION": "1.4.2",
-        "endpointURL": "/graphql",
+        "endpointURL": reverse("grapple_graphql"),
         "supports_subscriptions": has_channels,
     }
     if has_channels:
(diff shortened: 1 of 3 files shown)


grapple/urls.py
```diff
@@ -24,7 +25,9 @@ def graphiql(request):
 
 
 # Traditional URL routing
-urlpatterns = [url(r"^graphql", csrf_exempt(GraphQLView.as_view()))]
+urlpatterns = [
```
zerolab, line 28: suggestion: time to switch to `path`

otherwise, LGTM

ruisaraiva19: Updated!

grapple/urls.py
```diff
@@ -26,8 +25,8 @@ def graphiql(request):
 
 # Traditional URL routing
 urlpatterns = [
-    url(r"^graphql", csrf_exempt(GraphQLView.as_view()), name="grapple_graphql")
+    re_path(r"^graphql", csrf_exempt(GraphQLView.as_view()), name="grapple_graphql")
 ]
 
 if grapple_settings.EXPOSE_GRAPHIQL:
-    urlpatterns.append(url(r"^graphiql", graphiql, name="grapple_graphiql"))
+    urlpatterns.append(re_path(r"^graphiql", graphiql, name="grapple_graphiql"))
```
zerolab, line 32: it should not need `re_path`, `path("graphiql")` and `path("graphql")` should do it

ruisaraiva19: Updated!

zerolab (approved):

PR #195, opened 2021-11-01
feat(files): add missing fields to images and documents

### LINKED ISSUE #193: Add Image Focal point fields
Documentation says we can access focal point fields for an [ImageObjectType](https://wagtail-grapple.readthedocs.io/en/latest/general-usage/graphql-types.html#imageobjecttype). But these fields are not available at the APIs.

Looks like we need to add them at [BaseImageObjectType](https://github.com/GrappleGQL/wagtail-grapple/blob/e92abab64ea0dcacf68cd68ccd64a66c055e55f6/grapple/types/images.py#L17)?

ruisaraiva19:
- Closes #193

### DIFF
--- a/grapple/types/tags.py
+++ b/grapple/types/tags.py
@@ -3,13 +3,13 @@ import graphene
 from graphene_django.converter import convert_django_field
 from grapple.utils import resolve_queryset
 
-from modelcluster.contrib.taggit import ClusterTaggableManager
+from taggit.managers import TaggableManager
 from taggit.models import Tag
 
 from .structures import QuerySetList
 
 
-@convert_django_field.register(ClusterTaggableManager)
+@convert_django_field.register(TaggableManager)
 def convert_tag_manager_to_string(field, registry=None):
     return TagObjectType()
 
--- a/grapple/types/tags.py
+++ b/grapple/types/tags.py
@@ -3,6 +3,7 @@ import graphene
 from graphene_django.converter import convert_django_field
 from grapple.utils import resolve_queryset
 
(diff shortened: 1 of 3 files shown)


sks444 (approved):
This works well for me.

Thanks @ruisaraiva19!

grapple/types/tags.py
```diff
@@ -3,13 +3,13 @@
 from graphene_django.converter import convert_django_field
 from grapple.utils import resolve_queryset
 
-from modelcluster.contrib.taggit import ClusterTaggableManager
```
zerolab, line 6: question: what prompted this change?

ruisaraiva19: The `ClusterTaggableManager` extends the `TaggableManager` class, so by registering the base `TaggableManager` class we get proper values on the resolvers for all tag fields.

zerolab: 🤔 this'll probably fail on previews as modelcluster allow for that without objects being saved in the database

ruisaraiva19: Hum, that's a good point. Maybe we can register both managers?

zerolab: hmm,

tried 
```
@convert_django_field.register(TaggableManager)
@convert_django_field.register(ClusterTaggableManager)
def convert_tag_manager_to_string(field, registry=None):
    return TagObjectType(
```

locally and it doesn't explode. so let's go with that

ruisaraiva19: Updated!

PR #206, opened 2021-11-18
Support queryset slicing when there is a searchQuery

protasovse:


zerolab:
Hey @protasovse would you be so kind to add a test for this change?
Otherwise it looks like a useful addition

zerolab:
Added tests and merged in f4c2833. Thanks

PR #209, opened 2021-12-11
fix: docker build cannot find ../requirements

dopry:
- copy actual requirements to example folder instead of depending
   on chaining, the parent path isn't available in the docker build
   context.
- add missing wagtail-grapple requirement in requirements, I used the
   current version in pip, there is probably a better way to require it
   into the docker file.
- include channels requirements

I used this to push https://hub.docker.com/r/dopry/wagtail-grapple-example to meet my needs raised in https://github.com/GrappleGQL/wagtail-grapple/issues/208

### DIFF
--- a/example/Dockerfile
+++ b/example/Dockerfile
@@ -6,10 +6,12 @@ LABEL maintainer="hello@wagtail.io"
 ENV PYTHONUNBUFFERED 1
 ENV DJANGO_ENV dev
 
+COPY ./requirements-channels.txt /code/requirements-channels.txt
 COPY ./requirements.txt /code/requirements.txt
+
 RUN pip install --upgrade pip
 # Install any needed packages specified in requirements.txt
-RUN pip install -r /code/requirements.txt
+RUN pip install -r /code/requirements-channels.txt
 RUN pip install gunicorn
 
 # Copy the current directory contents into the container at /code/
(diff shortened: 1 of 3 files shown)


zerolab:
Hey @dopry,

Thank you for the time to make this PR. Unfortunately, cannot accept it as such because:

1. The example app is geared for local development and tests, therefore is not representative. I have plans to rejig the repository a bit in the coming weeks  to make its purpose more clear
2. `example/requirements.txt` install grapple as editable package in the state whatever branch you're working on. The changes would make local dev work cumbersome
3. With this PR, we'd have 2 x repeated requirements files that add to the maintenance overhead.

Having said that, I've extracted the example app into https://github.com/GrappleGQL/wagtail-grapple-demo, and reduced requirements to a single file. I believe this should do the job.

dopry:
I was concerned with those same issues when I was working on the PR. I knew it wasn't ideal for the local development flow, or repetition of work managing the requirements. Adding a second repo seems to me that it would add more maintenance overhead than managing two requirements files. A different approach came to mind after some consideration. How would you feel about me moving the Dockerfile up to the root of the repo, and adjusting it so it is working with the requirements.txt as they are?  It seems that might allow /example to work as intended in the local development workflow, and allow publishing of the example docker image from this same repo as a part of the same release workflow as the package itself...  thoughts?

PR #210, opened 2021-12-11
fix: docker build

dopry:
running `docker buiild .` in /example failed with an error that the
chained `../requirements.txt` could not be found. This commit moves
the Dockerfile to the root folder so the entire repo will be in the
default docker build context and includes the local wagtail-grapple
packaged in editable mode so the `example` site for any commit can be
easily published.


Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+# we're only going to copy the requirements for pip first, so if we're working
+# on the example site and testing docker builds we don't have to rebuid the
+# "pip requirements layers" every time we change a file.
+
+# copy local pip dependences
+COPY ./grapple /code/grapple
+COPY ./setup.cfg /code/setup.cfg
+COPY ./setup.py ./code/setup.py
+COPY ./requirements-channels.txt /code/requirements-channels.txt
+COPY ./requirements.txt /code/requirements.txt
```
fabienheureux, line 20: Suggestion: this step could be written like this
```suggestion
COPY requirements.txt setup.cf setup.py requirements-channels.txt requirements.txt /code/
```

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+# copy example site requires file
+COPY ./example/requirements.txt /code/example/requirements.txt
+COPY ./example/requirements-channels.txt /code/example/requirements-channels.txt
```
fabienheureux, line 24: ditto

dopry: If you wanna make that change go for it.  It isn't functionally different and makes it hard to see which file was added or removed on diffs, and also creates a single layer for all files so changing one invalidates layers for all.

zerolab:
@dopry I've merged this in 2eef039 (with the addition a big warning at the top of the file)

PR #217, opened 2022-02-25
Language code filter

### LINKED ISSUE #141: Filter on language
With Wagtail 2.11's built in support for multi-language, it would be very useful to have a way to filter pages based on language.

zerolab:
Fixes #141

### DIFF
--- a/grapple/types/pages.py
+++ b/grapple/types/pages.py
@@ -1,4 +1,5 @@
 import graphene
+from django.conf import settings
 from django.contrib.contenttypes.models import ContentType
 from django.dispatch import receiver
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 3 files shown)


grapple/types/pages.py
```diff
@@ -289,6 +299,10 @@ def resolve_pages(self, info, **kwargs):
                 else:
                     pages = pages.filter(content_type=ctype)
 
+            language_code = kwargs.pop("language_code", "")
```
kaedroho, line 302: Could we use the term "locale" here for consistency with the [API])https://docs.wagtail.org/en/stable/advanced_topics/api/v2/usage.html#filtering-pages-by-locale)?

PR #220, opened 2022-03-23
Add support for callables (properties and methods) within StreamFields

kbayliss:


example/home/blocks.py
```diff
@@ -89,6 +92,24 @@ class TextAndButtonsBlock(blocks.StructBlock):
...
+        GraphQLField("string_property", graphene.String, source="get_string_property"),
+        GraphQLField("string_method", graphene.String, source="get_string_method"),
+    ]
+
+    @property
+    def get_string_property(self, *args, **kwargs):
+        return "A simple string property."
+
+    def get_string_method(self, *args, **kwargs):
+        return "A simple string method."
```
zerolab, line 110: Most real-life use cases would use this in combination with some other local value, so it would be great to have this (e.g. return the `text` block value with a prefix, or perhaps with spaces removed)

grapple/actions.py
```diff
@@ -366,6 +366,13 @@ def streamfield_resolver(self, instance, info, **kwargs):
     return value
 
 
+def get_custom_resolver(cls, source):
+    if isinstance(getattr(type(cls()), source), property):
+        return lambda self, instance, info, **kwargs: getattr(cls(), source)
+    else:
+        return lambda self, instance, info, **kwargs: getattr(cls(), source)()
```
zerolab, line 373:
```suggestion
    klass = cls()
    if isinstance(getattr(type(klass), source), property):
        return lambda self, instance, info, **kwargs: getattr(klass, source)
    else:
        return lambda self, instance, info, **kwargs: getattr(klass, source)()
```

grapple/actions.py
```diff
@@ -391,15 +398,26 @@ class Meta:
 
     # Add any custom fields to node if they are defined.
     if hasattr(cls, "graphql_fields"):
-        for field in cls.graphql_fields:
-            if callable(field):
-                field = field()
+        for item in cls.graphql_fields:
+            has_custom_source: bool = hasattr(item, "field_source")
+
+            if all(
```
zerolab, line 404: praise: good sneaking in `all()` :)

I do wonder whether `if not has_custom_source and callable(item):` is more readable though

zerolab (changes requested):
A couple of minor changes, but looking good!

example/home/blocks.py
```diff
@@ -89,6 +92,24 @@ class TextAndButtonsBlock(blocks.StructBlock):
     ]
 
 
+@register_streamfield_block
+class TextWithCallableBlock(blocks.StructBlock):
+    text = blocks.CharBlock()
+
+    graphql_fields = [
+        GraphQLString("text"),
+        GraphQLField("string_property", graphene.String, source="get_string_property"),
```
zerolab, line 101: @kbayliss have you looked at the followin use case:

```python
class TextWithCallableBlock(blocks.StructBlock):
    text = blocks.CharBlock()

    graphql_fields = [
        GraphQLString("text"),
        GraphQLString("label"),
    ]

    @property
    def label(self):
        return "my block property"
```

? it would make it consistent with

```python
class MyPage(Page)
    #...
    graphql_fields = [
        GraphQLString("label"),
    ]

    @property
    def label(self):
        return "Event"
```

which "just works"

zerolab (approved):

PR #222, opened 2022-03-30
Pass instance values to callables

kbayliss:
Although [callables are now supported](https://github.com/GrappleGQL/wagtail-grapple/pull/220), you still cannot use any data held within the instance, which is arguably the main reason for wanting to use a callable.

This MR addresses that and passes all values for an instance to the callable method within a `StructBlock`. It also tidies a few things up.

### DIFF
--- a/docs/general-usage/model-types.rst
+++ b/docs/general-usage/model-types.rst
@@ -9,11 +9,39 @@ We created a number of built-in types for general use, but you can create your o
 using `Graphene <https://github.com/graphql-python/graphene/>`_ (Grapple's underlying library)
 and take advantage of Grapple's generic ``GraphQLField`` type.
 
+GraphQLField
+-------------
+.. module:: grapple.models
+.. class:: GraphQLField(field_name: str, field_type: type = None, required=None, **kwargs)
+
+    .. attribute:: field_name (str)
+
+        This is the name of the class property used in your model definition.
+
+    .. attribute:: field_type (type)
+
+        A Grapple model type such as ``GraphQLString`` or ``GraphQLBoolean``.
+
+    .. attribute:: required (bool=None)
+
+        Represents the field as non-nullable in the schema. This promises the client that it will have a value returned.
+
+    .. attribute:: kwargs
+
+        Useful keyword arguments:
+
+        * ``source`` (string)
+            You can pass a source string that is an attribute or method on the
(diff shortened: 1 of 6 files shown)


docs/general-usage/model-types.rst
```diff
@@ -9,11 +9,39 @@ We created a number of built-in types for general use, but you can create your o
...
+        Represents the field as non-nullable in the schema. This promises the client that it will have a value returned.
+
+    .. attribute:: kwargs
+
+        Useful keyword arguments:
+
+        * ``source`` (string)
+            You can pass a source string that is an attribute or method on the
+            class itself. If pointing to a method, all values from the instance
+            will be available within the `values` kwarg (`example <https://github.com/GrappleGQL/wagtail-grapple/blob/main/example/home/blocks.py#L99>`_).
```
zerolab, line 36: nitpick: if linking to code in the repo, use the commit hash as target may move.

having said that, this doc already includes code examples, so I suggest adding the snippet here

zerolab: further nitpick:

> If pointing to a method, all values from the instance will be available within the `values` kwarg

but only in a `StreamField` context. For `Page` or other models you only get access to the instance.

zerolab (changes requested):
The docs need tweaking.
The code changes LGTM. Kudos for the type hinting and extended tests!

docs/general-usage/model-types.rst
```diff
@@ -32,10 +33,26 @@ GraphQLField
 
         * ``source`` (string)
             You can pass a source string that is an attribute or method on the
-            class itself. If pointing to a method, all values from the instance
-            will be available within the `values` kwarg (`example <https://github.com/GrappleGQL/wagtail-grapple/blob/main/example/home/blocks.py#L99>`_).
+            class itself.
+
+            If used within a `StreamField`, the method will receive all values
+            from the instance via the `values` kwarg, e.g.:
```
zerolab, line 39: nitpick: not a kwarg 😂

kbayliss: We are passing this as a kwarg, no?

e.g.

```
return lambda self, instance, info, **kwargs: getattr(
    klass, graphql_field.field_source
)(values=get_all_field_values(instance=instance, cls=cls))
```

zerolab (approved):
LGTM

PR #232, opened 2022-07-20
Wagtail 4.0

kaedroho:


example/images/migrations/0002_alter_customimage_file_hash.py
```diff
@@ -0,0 +1,20 @@
+# Generated by Django 3.2.14 on 2022-07-25 13:13
```
zerolab, line 1: normally we keep the example app migrations to the min supported Wagtail version.
I double checked and with 2.15 only the blogpage body migration gets generated.

kbayliss:
https://github.com/torchbox/wagtail-grapple/blob/main/README.md?plain=1#L69 needs updating too

zerolab:
Addressed manually and merged in 52c36ce69b9b36 + parents

PR #233, opened 2022-07-20
Generic settings models

kaedroho:
Wagtail 4.0 adds generic settings models, this PR adds support for these into Grapple.

I've implemented this by adding generic settings into the existing "settings" endpoint. I've also added a "site" filter which will filter any site-specific settings to that given site (implementation based on https://github.com/wagtail/wagtail/commit/e6603260756350103a332f9585247dd0910604a3), and has no effect on generic settings.

I did it this way instead of a separate endpoint as this means the consumer of the API doesn't need to care about how the settings are configured in the back-end, just send the current site name and you'll be given the instance you need.

Depends on https://github.com/torchbox/wagtail-grapple/pull/232


zerolab (approved):
LGTM, with the caveats raised in #232

zerolab:
Merged in 154ef2391 + parents.

Thank you @kaedroho

Issue #234, opened 2022-07-21
`BASE_URL` needs replacing with `WAGTAILADMIN_BASE_URL` when using `wagtail>=3.0`


kbayliss:
From https://github.com/wagtail/wagtail/releases/tag/v3.0:

> Rename the setting BASE_URL (undocumented) to WAGTAILADMIN_BASE_URL and add to documentation, BASE_URL will be removed in a future release.

dopry:
I'm not sure that we necessarily want to replace all of those with WAGTAIL_ADMIN_BASEURL. That isn't necessarily the base URL for the site or the media. 

*  [grapple.utils.get_media_item_url](https://github.com/torchbox/wagtail-grapple/blob/8a54577e715ff4fedff3258b9efd37eae52036ae/grapple/utils.py#L184) 
   
   Media is not always hosted with the wagtail site. Most large sites I've worked on use cloud
   storage such as S3 and/or a Media CDN.  

   Media.url calls through AstractMedia.file -> FileField.url -> Storage.url

   django.core.files.storage.FileSystemStorage
   ```python 
   def url(self, name):
        if self.base_url is None:
            raise ValueError("This file is not accessible via a URL.")
        url = filepath_to_uri(name)
        if url is not None:
            url = url.lstrip('/')
        return urljoin(self.base_url, url)

   @cached_property
   def base_url(self):
        if self._base_url is not None and not self._base_url.endswith('/'):
            self._base_url += '/'
        return self._value_or_setting(self._base_url, settings.MEDIA_URL)
   ```

   I'm fairly certain the Storage implementation should be responsible for constructing a full URL here in nearly all cases, so we 
   shouldn't need the `if url[0] == "/":` branch. If we did enter it we should probably be preferencing MEDIA_URL to 
   WAGTAIL_ADMIN_BASEURL and checking the _base_url property of the relevant storage. 


* [grapple.types.streamfield.get_media_url](https://github.com/torchbox/wagtail-grapple/blob/4e9679b476327dfb366b8fd164db2e0b15ff1be5/grapple/types/streamfield.py#L343)
  
   this implementation of get_media_url is used solely used with get_embed_url in resolve_url. I'm not certain what object is being used here, but it should probably also favor MEDIA_URL and the base_url of the relevant storage. 

[...]

PR #242, opened 2022-08-07
fix: WAGTAIL_ADMIN_BASEURL warning in tests

dopry:
When running the test in the example the following warning is emitted.

```
WARNINGS:
?: (wagtailadmin.W003) The WAGTAILADMIN_BASE_URL setting is not defined
        HINT: This should be the base URL used to access the Wagtail admin site. Without this, URLs in notification emails will not display correctly.
```

applies to #234


zerolab (approved):
LGTM, thank you

PR #245, opened 2022-08-07
fix: cannot access the file because it is being used

### LINKED ISSUE #244: ERROR: test_properties_on_saved_example_document (example.tests.test_grapple.DocumentsTest)
When running the example tests for main, 44d9cec3766107c82556dc3d315c1ffe2f4dc999, I get the following error. 

```
======================================================================
ERROR: test_properties_on_saved_example_document (example.tests.test_grapple.DocumentsTest)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\...\wagtail-grapple\example\example\tests\test_grapple.py", line 828, in tearDown
    self.example_document.file.delete()
  File "C:\...\wagtail-grapple\example\venv\lib\site-packages\django\db\models\fields\files.py", line 107, in delete
    self.storage.delete(self.name)
  File "C:\...\wagtail-grapple\example\venv\lib\site-packages\django\core\files\storage.py", line 318, in delete
    os.remove(name)
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process: 'C:\\...\\wagtail-grapple\\example\\media\\documents\\example_6030QSD.txt'
```

dopry:
Windows will prevent a file from being deleted if a process
has an open handle on the file. The model needs to be deleted
to remove it's reference and then the file needs to be closed
before removing.

Fixes #244

### DIFF
--- a/example/example/tests/test_grapple.py
+++ b/example/example/tests/test_grapple.py
@@ -706,10 +706,9 @@ class DocumentsTest(BaseGrappleTest):
 
         self.assertEqual(example_doc.id, 1)
         self.assertEqual(example_doc.title, "Example File")
-
-        example_doc.file.seek(0)
-
-        self.assertEqual(example_doc.file.readline(), b"Hello world!")
+        with example_doc.open_file() as file:
+            file.seek(0)
+            self.assertEqual(file.readline(), b"Hello world!")
 
         self.assertNotEqual(example_doc.file_hash, "")
         self.assertNotEqual(example_doc.file_size, None)


example/example/tests/test_grapple.py
```diff
@@ -825,7 +825,10 @@ def test_query_url_field_with_direct_document_serve_method(self):
         settings.WAGTAILDOCS_SERVE_METHOD = serve_method_at_test_start
 
     def tearDown(self):
-        self.example_document.file.delete()
+        file = self.example_document.file
+        self.example_document.delete()
+        file.close()
+        file.delete(save=False)
```
zerolab, line 831: could this be reduced to:

```
file = self.example_document.file
file.close()
file.delete()
```
I don't use Windows, so cannot test.

dopry: I tried and that didn't work. But while verifying I was getting in consistent results and looked more closely at the test. I found the file was being accessed without a context manager, so I wrapped the file access in a context manager instead.

zerolab (approved):
LGTM. thank you

PR #247, opened 2022-08-07
fix: Unlink of example/db.sqlite3 failed

### LINKED ISSUE #246: Cannot switch branches because of example/db.sqllite3
When I have the server running and start switching branches I get the error
```
"git" checkout --force "main"
Unlink of file 'example/db.sqlite3' failed. Should I try again? (y/n)
```

this is a pretty common activity for me as a I switch between test branches and don't really want to alter the database... plus the first instruction in the getting started is to delete the database... Unless it serves some other purpose, it seems like it should be removed.

dopry:
The second instruction for developers is to remove this file. It creates
issues when switching branches during development with the server
running. This commit removes it. If standard test data is needed, the
project should use factories or fixtures instead of a backed in DB for
better cross platform compatibilty and consistency.

Fixes #246

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,6 @@ jobs:
       - name: Run Tests
         run: |
           cd example
-          rm db.sqlite3
           python manage.py migrate
           python manage.py test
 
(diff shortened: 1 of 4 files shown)


zerolab (approved):
Good shout on this. I had a WIP branch with various tidy ups (#251) that was doing it, but let's get yours in.

Thank you

Issue #249, opened 2022-08-07
Pages Query multiple content types


dopry:
It would be nice to query more than one content type at a time. We have a few page types and would like to collate some of them into a single view.

PR #250, opened 2022-08-07
feat: comma separated list of content_type

dopry:
applies to #249

### DIFF
--- a/grapple/types/pages.py
+++ b/grapple/types/pages.py
@@ -1,5 +1,6 @@
 import graphene
 from django.contrib.contenttypes.models import ContentType
+from django.db.models import Q
 from django.dispatch import receiver
 from django.utils.translation import gettext_lazy as _
 from graphene_django.types import DjangoObjectType
@@ -233,7 +234,9 @@ def PagesQuery():
             graphene.NonNull(lambda: PageInterface),
             content_type=graphene.Argument(
                 graphene.String,
-                description=_("Filter by content type. Uses the `app.Model` notation."),
+                description=_(
+                    "Filter by content type. Uses the `app.Model` notation. Accepts a comma separated list of content types."
+                ),
             ),
             in_site=graphene.Argument(
                 graphene.Boolean,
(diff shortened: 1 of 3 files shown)


grapple/types/pages.py
```diff
@@ -284,10 +286,17 @@ def resolve_pages(self, info, **kwargs):
...
+                app_labels = []
+                models = []
+                for content_type in content_types:
+                    app_label, model = content_type.strip().lower().split(".")
+                    app_labels.append(app_label)
+                    models.append(model)
                 pages = pages.filter(
-                    content_type__app_label=app_label, content_type__model=model
+                    content_type__app_label__in=app_labels,
+                    content_type__model__in=models,
```
zerolab, line 299: This doesn't feel right, we need to check pairs of app label/model, rather than two lists. suggest constructing a set of OR-ed `Q()` objects

zerolab (changes requested):

grapple/types/pages.py
dopry: good catch.  Updated and rebased.

zerolab (commented):
The changes look good to me now.

I do feel we need to update the docs https://wagtail-grapple.readthedocs.io/en/latest/general-usage/graphql-types.html?highlight=contentType#pageinterface to mention this, otherwise it stays hidden

dopry:
I would argue that it is already hidden. In the docs, contentType only says String. It doesn't indicate that you need to use app.Model notation. I don't necessarily I want to get into writing a lot of docs that is not my forte. I did update the description so it is apparent when working in graphiql.

dopry:
I added a quick comment in the doc...  I also updated the docs for my ancestor/parent PR. It would be nice if the type docs were just derived from the graphql schema so contributors didn't need to update them in two places...

zerolab (approved):

zerolab:
> It would be nice if the type docs were just derived from the graphql schema so contributors didn't need to update them in two places

I agree. It is always a trade-off between working on a feature or working on thing around (like docs, tooling, etc), and the time available. If you have the time and want to tinker, please do :)

dopry:
Thanks for getting this merged so quickly! I really appreciate it. I, unfortunately, don't have time to pursue the documentation generation endeavor. I need to be focused on my current project for the next month or two and have other open source projects that consume my leisure time. I do expect you'll be seeing me around the grapple/headless-preview/admin client space for the duration of this project.

PR #251, opened 2022-08-08
Various tidy ups

zerolab:
- GitHub Actions versions
 - Update various linters
 - Moves build and publish to PyPI to own workflow, with trigger from release creation, rather than just pushing a tag
 - drops the example sqlited db (related #247)


kaedroho (approved):
All look good to me, thanks!

PR #252, opened 2022-08-09
fix: do not alter embed urls

dopry:
in general embeds are fully qualified urls and the embed finder
is largely responsible for verifying that. If a finder allowed a
relative_url, generating a fully qualified url should be delegated
to the finder and not something that grapple should be trying to
do. Otherwise we are violating the single responsility prinicipal
and seperation of concerns.

applies to #234


zerolab (approved):
Nice little tidy up, thanks @dopry

Issue #204, opened 2021-11-13
`RichTextField` returns internal markup while `RichTextBlock` returns browser-ready HTML


isolationism:
First, a hat tip to all the developers working on this project; thank you for all you have done.

I encountered this one today. The issue can be described thus:

```python
class ExamplePage(HeadlessPreviewMixin, Page):
    body = RichTextField()
    stream_body = StreamField([
        ('paragraph', RichTextBlock()),
        ...
    ])
    graphql_fields = [
        GraphQLString("body"),
        GraphQLStreamfield("stream_body"),
    ]
```

In the result of my GraphQL query for the `ExamplePage` object described above, the `body` field will be in Wagtail's [Rich Text Internals](https://docs.wagtail.io/en/stable/extending/rich_text_internals.html#rich-text-internals) format, which allows the Gatsby client to interpolate objects like images correctly. This is good!

Unfortunately, the `stream_body` field is returned in browser-ready HTML, which is not so good, at least for my purposes — ideally I'd like the resolver to return the actual `raw_value` for `RichText` objects as the property name suggests. The current implementation returns what I would describe as a rendered or interpolated value.

I'm not sure what the expected behaviour _should_ be, but I was able to patch the issue trivially by adding an additional condition to `types.streamfield.StreamFieldInterface.resolve_raw_value()` as follows:

```python
    def resolve_raw_value(self, info, **kwargs):
        if isinstance(self, blocks.StructValue):
            # This is the value for a nested StructBlock defined via GraphQLStreamfield
            return serialize_struct_obj(self)
        elif isinstance(self.value, dict):
            return serialize_struct_obj(self.value)
        # per https://docs.wagtail.io/en/stable/extending/rich_text_internals.html#data-format
        # RichTextBlock.value is converted to browser HTML; this change returns the "internal source"
        # so we can parse and mark-up client-side using the same logic as for `RichTextField`
        elif isinstance(self.value, RichText):
            return self.value.source


        return self.value
```

I'm not sure if this is the desired behaviour or not, but this would amount to a breaking change if existing code expects to carry forward the current, heterogeneous rendering approach between these two similar objects.

Can anyone from the project chime in? I'd be happy to submit a PR with the code above if the maintainers consider this a defect.

zerolab:
@isolationism thank you for the detailed issue description.

There are two things here:
1. `RichTextField` and `RichTextBlock` produce inconsistent results
2. The output of these fields is subject to individual needs (e.g. see #200)

[...]

Issue #200, opened 2021-11-09
Correct GraphQL field for wagtail RichTextField


sks444:
```python
from wagtail.core.fields import RichTextField

@register_snippet
class TestSnippet(models.Model):
    title = models.CharField(max_length=50)
    description = RichTextField(
        blank=True
    )

    panels = [
        FieldPanel('title'),
        FieldPanel('description'),
    ]

    graphql_fields = [
        GraphQLString('title', required=True),
        GraphQLString('description', required=True),
    ]
```

I have been using `GraphQLString` for wagtail `RichTextField` which doesn't seem to be the most accurate way..as Richtext field output should ideally be passed through richtext method for the rendered content. E.g. In template there is richtext filter and [expand_db_html](https://github.com/wagtail/wagtail/blob/571b9e1918861bcbe64a9005a4dc65f3a1fe7a15/wagtail/core/rich_text/__init__.py#L24) method.

One example is when using apostrophe `'`  inside richtext fields, it returns code instead of the string. 

```shell
In [62]: object.text
Out[62]: '<p data-block-key="t4jac">dc csdc &#x27;This is the string&#x27;</p>'

```

sks444:
Thanks @zerolab,

Do we want this to be handled in `GraphQLString` itself?

For testing purposes I added this bit of code to the [model_resolver](https://github.com/GrappleGQL/wagtail-grapple/blob/554cb2d5d253bffd4213fbfb4fb751351ccc7a29/grapple/actions.py?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L170)..

```python
        cls_field = getattr(instance, field.field_source)
        if field.field_type.of_type == graphene.String:
            cls_field = render_to_string(
                "wagtailcore/richtext.html",
                {"html": expand_db_html(cls_field)}
            )
```

It still returns the apostrophe `'` as code `&#x27;` in the API 🤔

zerolab:
@sks444 this would need to be handled withing `GraphQLString`. 
Worth coordinating with #204. Need to set aside some time to look into the additional encoding.

timmysmalls:
Or a GraphQLRichText model? I'm not sure if what's the most elegant way to go about it? I could implement either and create a MR.

zerolab:
[...]

PR #255, opened 2022-08-11
Render expanded HTML for RichTextFields

Tijani-Dia:
This PR aims to solve the issue mentioned in #200 and #204.


grapple/actions.py
```diff
@@ -227,6 +229,17 @@ def get_nested_field(cls, extract_key):
         if callable(cls_field):
             return cls_field(info, **kwargs)
 
+        # Expand HTML if the value's field is richtext
```
kaedroho, line 232: Would this affect existing installations?

If so, is there any way we can make this opt-in, so that sites won't break on upgrade?

zerolab: This is on my list to review on Friday. it will most certainly affect existing installations

https://github.com/torchbox/wagtail-grapple/issues/204#issuecomment-968283281 mentions

> I think the better option woul be to add a setting that controls the output on both `RichTextField` and `RichTextBlock`

zerolab:
Superseded by #259. Thanks @Tijani-Dia

PR #256, opened 2022-08-15
Make SnippetChooserBlock return proper snippet values

jams2:
This PR updates the `snippet` field of the grapple `SnippetChooserBlock` class so snippet values in stream fields return the snippet model's fields, rather than just a single string field.

To achieve this both the grapple `SnippetChooserBlock` and `SnippetsQuery.Mixin` classes need access to the `SnippetObjectType`, which depends on the snippets registry being populated. I have added the `SnippetTypes` class to `grapple.types.snippets` for the following purposes:
- not creating the `SnippetObjectType` class when a project has no snippets (which causes an error, as a `graphene.Union` subclass requires at least one type);
- creating the `SnippetObjectType` class lazily (i.e. not until `grapple.types.snippets.SnippetsQuery.Mixin` or `grapple.types.streamfield.register_streamfield_blocks.SnippetChooserBlock` are required, and the snippets registry has been populated)


example/home/test/test_blog.py
```diff
@@ -464,6 +469,60 @@ def test_blog_body_pagechooserblock(self):
...
+                text
+            }
+        }
+        """
+        query_blocks = self.get_blocks_from_body(
+            block_type, block_query=block_query, page_id=blog_page.id
+        )
+        self.assertEqual(len(query_blocks), 1)
+        block = query_blocks[0]
+        self.assertEqual(block["blockType"], block_type)
```
ababic, line 495: Question: Would we expect `get_blocks_from_body()` to ever return anything but `SnippetChooserBlock` value types when that is supplied as the first argument? I can fully understand the value of the following two assertions, but this one, I'm not so sure about.

jams2: Yep, I think you're correct, it's unnecessary.

zerolab:
@jams2 would be great to fix the CI failures ahead of Friday (when I be doing a concerted grapple review/maintenance effort)

zerolab:
Merged in 60bf85a

PR #257, opened 2022-08-16
Support specifying format in image srcSet query

### LINKED ISSUE #238: Add a way to specify image format when generating `src` set of renditions
We can specify different image formats when using [`srcset`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/srcset). However, the `srcSet` being currently generated for an image can only contain the default image format.

It would be nice to have a way to specify different formats:
```
{
  images {
    ... on CustomImage {
      title
      srcSet(sizes: [360, 1024], format: "webp")
  }
}
```

jams2:
Closes #238 

Allows the format of renditions to be specified when performing an Image srcSet query

### DIFF
--- a/docs/general-usage/graphql-types.rst
+++ b/docs/general-usage/graphql-types.rst
@@ -122,7 +122,10 @@ need for Gatsby Image features to work (see Handy Fragments page for more info):
     fileHash: String!
     renditions: [ImageRenditionObjectType]
     src: String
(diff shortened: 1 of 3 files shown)


zerolab (approved):
Lovely stuff.
Note to self: add note about the need to add the format to `ALLOWED_IMAGE_FILTERS`, if used

PR #258, opened 2022-08-18
Multi site support

kaedroho:
Headless sites usually wouldn't be hosting the API on the same hostname as the frontend. It may also be hosting multiple frontend domains off the same API domain.

This PR enables the frontend to filter pages, and sites by a particular site hostname.

### DIFF
--- a/grapple/types/sites.py
+++ b/grapple/types/sites.py
@@ -2,6 +2,7 @@ import graphene
 from django.contrib.contenttypes.models import ContentType
 from django.utils.translation import gettext_lazy as _
 from graphene_django.types import DjangoObjectType
+from graphql.error import GraphQLError
 
 try:
     from wagtail.models import Page as WagtailPage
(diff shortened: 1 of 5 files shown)


zerolab (approved):
Nice addition. Thank you @kaedroho 

rebasing manually and will add `get_site_filter` to the page subscription query too

grapple/types/sites.py
```diff
@@ -82,12 +83,16 @@ def resolve_sites(self, info, **kwargs):
...
 
-            try:
-                if id:
-                    return Site.objects.get(pk=id)
-                elif hostname:
-                    return Site.objects.get(hostname=hostname)
-            except BaseException:
-                return None
+            if id:
+                return Site.objects.filter(pk=id)
```
zerolab, line 87: note: this should be `.get()`, otherwise we get a `Received incompatible instance \"<QuerySet [<Site: grapple.test:8000 [default]>]>` error

zerolab:
Merged in bcf83c88ccbcd1e17e61b1e07babf98604463b7f and parents.

grapple/types/sites.py
kaedroho: Oops, I forgot to add `.first()` here. I used this instead of `.get()` so that it returns `None` if the site is not found (replicating what `except BaseException:` was doing before)

PR #261, opened 2022-08-24
Capture FieldDoesNotExist errors when getting the type of a model instance's field

Tijani-Dia:
The goal of this PR is to avoid `FieldDoesNotExist` errors for GraphQL fields that aren't actual fields in the model.
I've added a failing test to exhibit the issue.


zerolab (approved):
LGTM

PR #263, opened 2022-08-31
Avoid `N*M + 1` problem on image queries

Tijani-Dia:
Currently, this only contains a test to show the problem mentioned in #264.


zerolab:
#271 builds on this, so closing

Issue #265, opened 2022-09-02
Document `CustomQuery` example for `register_schema_query` hook


Morsey187:
I'm on a project where we've found ourselves having to use a custom query via the `register_schema_query` hook for a snippet model, we're no longer querying via snippets because of this and I've had to resolve a RichtextField myself, for example:

```python
class AlertObjectType(graphene.ObjectType):
    title = graphene.String(required=True)
    richtext = graphene.String(required=True)

    def resolve_message(self, info, **kwargs):
        if grapple_settings.RICHTEXT_FORMAT == "html":
            return render_to_string(
                "wagtailcore/richtext.html",
                {"html": expand_db_html(self.richtext)},
            )
        return self.richtext
```
We could either document an example of a `CustomQuery` class with a wagtail field on this [page](https://wagtail-grapple.readthedocs.io/en/latest/general-usage/hooks.html?highlight=customquery#query) to hightlight this issue, or create a fix and document a cleaner solution.

zerolab:
Improving docs is a always a good thing.

Alternatively, we could perhaps create a new `GraphQLRichText` model that does all of this for devs?

Morsey187:
I'd be in favour of adding a new GraphQLRichText field type. I'm on holiday for 2 weeks now but definitely keen to pick this one up myself if no one gets to it before then.

zerolab:
This is now fixed by #266

PR #266, opened 2022-09-08
Add GraphQLRichText and RichText(Scalar) types

jams2:
Adds a `GraphQLRichText` type, as suggested in https://github.com/torchbox/wagtail-grapple/issues/265. There are a couple of small points I'm unsure about, I will address them in comments below.


grapple/types/rich_text.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class RichText(Scalar):
+    @staticmethod
+    def serialize(rich_text: str):
+        if grapple_settings.RICHTEXT_FORMAT == "html":
+            return render_to_string(
+                "wagtailcore/richtext.html", context={"html": expand_db_html(rich_text)}
+            )
+        return rich_text
```
jams2, line 19: I have not implemented `parse_literal` or `parse_value` (see https://docs.graphene-python.org/en/stable/types/scalars/#custom-scalars). If I understand correctly these are needed to make the type writable. Should they be implemented?

grapple/actions.py
```diff
@@ -230,24 +230,6 @@ def get_nested_field(cls, extract_key):
...
-        else:
-            if (
-                type(field_model) is RichTextField
-                and grapple_settings.RICHTEXT_FORMAT == "html"
-            ):
-                return render_to_string(
-                    "wagtailcore/richtext.html", {"html": expand_db_html(cls_field)}
-                )
-
         # If none of those then just return field
```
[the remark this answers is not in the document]
jams2: Added this back with tests in https://github.com/torchbox/wagtail-grapple/pull/266/commits/62739133dce84fda16568676cfec5cae78c0b0d1

grapple/types/rich_text.py
jams2: Also related to this method: the `== "html"` case could be simplified to `WagtailRichText(rich_text).__html__()`, however grapple is using its own `richtext.html` template.

The grapple template was introduced in https://github.com/torchbox/wagtail-grapple/pull/90, to avoid adding `<div class="rich-text">...</div>` to the output, which was included in the Wagtail `richtext.html` template at the time.

The wrapping div was removed from Wagtail for 2.10 (see https://github.com/wagtail/wagtail/commit/a1a2c35c1c1e78efa38bed0074ae0c7968770e84). However, grapple's `richtext.html` contains a trailing newline, where Wagtail's does not.

Should we remove the template from grapple, and use wagtail's? If so, should I open a new ticket for it or make the change in this PR?

grapple/types/rich_text.py
```diff
@@ -0,0 +1,25 @@
...
+
+    parse_value = str
+
+    @staticmethod
+    def parse_literal(node):
+        return str(node.value)
```
jams2, line 25: I'm not sure what the action is that causes this method to be called. Graphene and graphql-core docs are pretty vague on this and I couldn't trigger through a create or update mutation. Perhaps my lack of knowledge on the implementation of graphql servers.

grapple/types/rich_text.py
zerolab: I would be all up for this, however `wagtail.contrib.legacy.richtext` gives me pause.
saying that, it should not be an issue on headless sites, so let's do the requisite tidy up and add a good nice notice for upgrades

grapple/types/streamfield.py
```diff
@@ -304,12 +303,7 @@ class Meta:
         interfaces = (StreamFieldInterface,)
 
     def resolve_value(self, info, **kwargs):
-        # Allow custom markup for RichText
-        if grapple_settings.RICHTEXT_FORMAT == "html":
-            return render_to_string(
-                "wagtailcore/richtext.html", {"html": expand_db_html(self.value.source)}
-            )
-        return self.value.source
+        return RichTextType.serialize(self.value.source)
```
zerolab, line 306: was going to suggest that we change the definition of the `RichTextBlock` object type
to have `value = RichTextType(required=True)`, but value is a dict in the form of `{"source": "raw value"}`, so that would've needed handling ing the `RichText` scalar definition which is not as neat. so 👍

docs/getting-started/examples.rst
```diff
@@ -14,6 +14,7 @@ Wagtail docs:
 .. code-block:: python
```
zerolab, line 14: `GraphQLRichText` needs documenting in `docs/general-usage/model-types.rst`

zerolab (changes requested):
Lots of great stuff in here, @jams2.

Needs a few changes, as outlined in the review.

docs/getting-started/examples.rst
jams2: Updated in https://github.com/torchbox/wagtail-grapple/pull/266/commits/763c244a8b659337f077672a1cb6c51ee70d6b98

grapple/types/rich_text.py
jams2: Good thinking, updated in https://github.com/torchbox/wagtail-grapple/pull/266/commits/98ae8c77d0b0d766323833c1bb724f1b81e0cc41

example/home/test/test_advert.py
```diff
@@ -68,3 +78,32 @@ def test_advert_single_query_required(self):
...
+
+        # Declared with GraphQLString, default field source
+        self.assertEqual(advert["extraRichText"], self.richtext_sample_rendered)
+
+        with override_settings(GRAPPLE={"RICHTEXT_FORMAT": "raw"}):
+            executed = self.client.execute(query, variables={"url": self.advert.url})
+            advert = executed["data"]["advert"]
+            self.assertEqual(advert["richText"], self.richtext_sample)
+            self.assertEqual(advert["stringRichText"], self.richtext_sample)
+            self.assertEqual(advert["extraRichText"], self.richtext_sample)
```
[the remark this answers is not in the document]
jams2: Added tests in https://github.com/torchbox/wagtail-grapple/pull/266/commits/170b15ea8893e5a4fe260a87a885a3d009ab7413

example/home/models.py
```diff
@@ -235,18 +239,32 @@ class Author(Orderable):
...
     text = models.CharField(max_length=255)
+    rich_text = RichTextField(blank=True, default="")
+    extra_rich_text = RichTextField(blank=True, default="")
+
+    panels = [
+        FieldPanel("url"),
+        FieldPanel("text"),
+        RichTextFieldPanel("rich_text")
+        if settings.WAGTAIL_VERSION < (3, 0)
+        else FieldPanel("rich_text"),
```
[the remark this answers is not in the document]
jams2: Updated in https://github.com/torchbox/wagtail-grapple/pull/266/commits/170b15ea8893e5a4fe260a87a885a3d009ab7413

zerolab (approved):

PR #269, opened 2022-09-21
Fix returned ID for nested StructBlock in StreamBlock

### LINKED ISSUE #268: Querying the ID for blocks in a nested StructBlock inside a StreamBlock returns incorrect value
We've had an issue querying the IDs for blocks in a nested `StructBlock` inside a `StreamBlock`.

I've been to reproduce the issue using the [`CarouselBlock`](https://github.com/torchbox/wagtail-grapple/blob/main/example/home/blocks.py#L63) in the `example` directory. See 8b0f547941d5468eca43b87454ee2281adb7189d.

Instead of the ID, the query returns the block name inside the parent StreamBlock. So in the example in the commit, it would return the following:
```
[{
    'blockType': 'CarouselBlock',
    'blocks': [
        {
            'id': 'image',
            'image': {'src': 'http://localhost:8000/media/original_images/example_QsYB5i6.jpg'}
        },
        {
            'id': 'image',
            'image': {'src': 'http://localhost:8000/media/original_images/example_n1Sof73.jpg'}
        }
    ]
}]
```

It's worth noting that this caused us an obscure bug, where only the first block of each StructBlock inside the StreamBlock was being rendered. (This is potentially due to GraphQL client servers caching objects by their ID.)

Tijani-Dia:
Fixes #268.

The issue was with `StructBlockItem` being passed the wrong arguments.

_Note_: It'd be nice to start using the [`blocks_by_name`](https://docs.wagtail.org/en/latest/topics/streamfield.html#retrieving-blocks-by-name) method in tests where we retrieve a block by its name when Wagtail 4 support is added.

### DIFF
--- a/example/home/test/test_blog.py
+++ b/example/home/test/test_blog.py
@@ -735,6 +735,27 @@ class BlogTest(BaseGrappleTest):
                 self.assertEquals(button["buttonText"], "Take me to the source")
                 self.assertEquals(button["buttonLink"], "https://wagtail.io/")
 
+    def test_nested_structvalue_block_id(self):
+        block_type = "CarouselBlock"
+        query_blocks = self.get_blocks_from_body(
+            block_type,
+            block_query="""
+                blocks {
+                    ...on ImageChooserBlock {
+                        id
+                    }
+                }
+            """,
+        )
+
+        blocks = query_blocks[0]["blocks"]
+
+        # Check that the id returned matches the original block's ID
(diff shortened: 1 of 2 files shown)


zerolab:
@Tijani-Dia we support Wagtail. I think the key bit here is to start using it in earnest when we don't support anything < 4.0

zerolab (approved):

PR #271, opened 2022-10-29
Prefetch renditions

### LINKED ISSUE #264: `N*M + 1` problem when fetching image renditions
When we fetch multiple image renditions using the `srcSet` parameter on image queries, we end up having an `N * M + 1` situation, where `N` is the total number of images to retrieve renditions for and `M` is the number of renditions to generate.

I've opened a draft PR with a test which confirms this behaviour here: #263.

zerolab:
Only works with Wagtail 3+

Fixes #264 and replaces #263 
@Tijani-Dia care to have a look?

### DIFF
--- a/example/example/tests/test_grapple.py
+++ b/example/example/tests/test_grapple.py
@@ -1018,8 +1018,6 @@ class ImagesTest(BaseGrappleTest):
     def test_src_set_num_queries(self):
         sizes = [360, 720, 1024]
         filters = [f"width-{size}" for size in sizes]
-        N = 5  # Number of images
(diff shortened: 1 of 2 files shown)


zerolab:
Will be merging this later this afternoon if no shouts until then

Tijani-Dia:
Hey Dan, thank you for this. I'm wondering if we could specify the filters the user requested when we use the  `prefetch_renditions` method. Otherwise, that looks good to me.

zerolab:
That would be ideal scenario. Needs more digging, but I need to make a release so will merge this as a first step improvement

Issue #275, opened 2022-11-28
GetPage(id: ..., token: "...") returning null for unpublished pages


dopry:
My headless preview call the GraphQL API to get the page data with the token. When pages are unpublished it is returning null, so that pages cannot be previewed before publishing. 

Given
```
{
  page( id: 367, token: "id=367:1ozhYP:6r6kIgWON9Lag-99mKOs83UioM8T9c2-z93m2cut0W4" ) {
    title
  }
}
```

When the page is unpublished the response is  ```{  "data":{ "page": null }}  ```
When the page is published the response is ```{ "data": {  "page": {  "title": "Test Preview unpublished"  }}}```

dopry:
I'm feel like the problem is https://github.com/torchbox/wagtail-grapple/blob/bcf83c88ccbcd1e17e61b1e07babf98604463b7f/grapple/types/pages.py#L181

I feel like live() should not be included if token is set.

zerolab:
Good 🕵️‍♂️!

dopry:
@zerolab I'll work on a fix to get_specific_page. I feel like, in a future release, it may be wise to have a separate graphql query for getPagePreview(token: String!). There are a lot of conditions stacking up in get_specific_page, and making .live() and .public() conditional could lead to errors where we accidentally expose draft content to the public. 

the token has the page id embedded.  So a get_page_preview implementation could potentially be simplified to something like
```python
def get_page_preview(token=None):
   id_str = token.split(':")[0]
   id = id_str.splt("=")  
   page = WagtailPage.objects.get(pk=id).specific()
   if page:
        page_type = type(page)
        if hasattr(page_type, "get_page_from_preview_token"):
            preview = page_type.get_page_from_preview_token(token)
   if preview:
        return preview
```

PR #277, opened 2022-11-29
fix: previews for unpublished pages

dopry:
applies to #275 

I couldn't find any tests for the token calls and don't have a clue where to start generating them.

### DIFF
--- a/grapple/types/pages.py
+++ b/grapple/types/pages.py
@@ -169,6 +169,66 @@ class Page(DjangoObjectType):
         interfaces = (PageInterface,)
 
 
+def get_preview_page(token):
+    """
+    Get a preview page from a token.
+    """
+    try:
+        token_params_list = token.split(":").pop(0)
+        token_params_kvstr = token_params_list.split(";")
+
+        params = {}
+        for arg in token_params_kvstr:
+            key, value = arg.split("=")
+            params[key] = value
+
+        id = params.get("id")
+        if id:
+            """
+            This is a page that had already been saved. Lookup the class and call get_page_from_preview_token.
+
+            TODO: update headless preview to always send page_type in the token so we can always
+            the if content_type branch and elimiate the if id branch.
+            """
+            page = WagtailPage.objects.get(pk=id).specific
+            if page:
+                cls = type(page)
+                """
+                get_page_from_preview_token is added by wagtail-headless-preview,
+                this condition checks that headless-preview is installed and enabled
+                for the model.
+                """
+                if hasattr(cls, "get_page_from_preview_token"):
+                    """we assume that get_page_from_preview_token validates the token"""
+                    return cls.get_page_from_preview_token(token)
+
+        content_type = params.get("page_type")
+        if content_type:
+            """
+            this is a page which has not been saved yet. lookup the content_type to get the class
+            and call get_page_from_preview_token.
+            """
+            app_label, model = content_type.lower().split(".")
+            ctype = ContentType.objects.get(app_label=app_label, model=model)
+            if ctype:
+                cls = ctype.model_class()
+                """
+                get_page_from_preview_token is added by wagtail-headless-preview,
+                this condition checks that headless-preview is installed and enabled
+                for the model.
+                """
+                if hasattr(cls, "get_page_from_preview_token"):
+                    """we assume that get_page_from_preview_token validates the token"""
+                    return cls.get_page_from_preview_token(token)
+    except Exception:
+        """
+        catch and suppress errors. we don't want to expose any information about unpublished content
+        accidentally.
+        TODO: consider logging here.
+        """
+        return None
+
+
 def get_specific_page(
     id=None, slug=None, url_path=None, token=None, content_type=None, site=None
 ):
@@ -177,9 +237,13 @@ def get_specific_page(
     """
     page = None
     try:
+        if token:
+            return get_preview_page(token)
+
         # Everything but the special RootPage
         qs = WagtailPage.objects.live().public().filter(depth__gt=1).specific()
         ctype = None
+
         if site:
             qs = qs.in_site(site)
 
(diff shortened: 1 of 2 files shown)


dopry:
I am testing with the following graphql query,  we'll need to setup some sort of similar test scenario.  note in both scenarios my frontend preview code has to copy the id or page_type out of the token to pass to the graphql query. I feel like maybe I should go deeper and just have the token take precendence over everything else if it is present. 

```
{
  previewUnpublishedSavedDraft: page(
    id: 367
    token: "id=367:1p05oi:jDIi4YDGXs-6YgoOVz05HgyWrEwnOEoP3PTsEEF8RD4"
  ) {
    title
  }
  
  previewUnsaved: page(
    contentType: "typenetwork.TnPage"
    token: "parent_id=367;page_type=typenetwork.TnPage:1p05wC:jJsi2R0bD_IOHDMbZ4kKc34lH2myylKY_vGQ7_T_Hlk"
  ) {
    title
  }
  
  page(id: 367) {
    title
  }
}
```

dopry:
I force pushed an update where token takes priority over all other args is present and the params required for fulfillment of a token query are extracted from the token. 

My update test graphql looks more like 
```
{
  "data": {
    "previewUnpublishedSavedDraft": {
      "title": "Not Published"
    },
    "previewUnsaved": {
      "title": "Title"
    },
    "page": null
  }
}
```

I still need help figuring out how to set the test scenarios for 2 token types (saved and unpublished, unsaved and unpublished), all the published scenarios should be covered by existing tests.  We may also want to add tests to ensure unpublished content isn't accessible without the token. I'm not sure that those tests should necessarily hold up this PR since we don't already have coverage on those feature and this is a bug that could be impacting production sites.

grapple/types/pages.py
```diff
@@ -169,6 +169,66 @@ class Meta:
...
+    """
+
+    """ 
+    We get id when previewing a page that has been saved. We're going to use it
+    to get the page so we can lookup it's class in order to call 
+    get_page_from_preview_token
+    """
+    id = params.get("id")
+    if id:
+        page = WagtailPage.objects.get(pk=id).specific
```
zerolab, line 196: please wrap this and the ctype one in a try/except `<model>.DoesNotExist`

grapple/types/pages.py
```diff
@@ -169,6 +169,66 @@ class Meta:
...
+
+    params = {}
+    for arg in token_params_kvstr:
+        key, value = arg.split("=")
+        params[key] = value
+
+    """
+    Is there a more convenient way to get at the get_page_from_preview_token
+    method? 
+    """
```
zerolab, line 187: Not unless we extract https://github.com/torchbox/wagtail-headless-preview/blob/main/src/wagtail_headless_preview/models.py#L61-L72 into its own method

grapple/types/pages.py
```diff
@@ -169,6 +169,66 @@ class Meta:
...
+    id = params.get("id")
+    if id:
+        page = WagtailPage.objects.get(pk=id).specific
+        if page:
+            cls = type(page)
+            """ 
+            is there a particular version of wagtail that requires checking that
+            get_page_from_preview_token exists? is there a version of wagtail 
+            at which we can drop this conditional safely?
```
zerolab, line 202: `get_page_from_preview_token` comes for wagtail-headless-preview and is Wagtail-versiona agnostic

grapple/types/pages.py
```diff
@@ -169,6 +169,66 @@ class Meta:
...
+            """
+            if hasattr(cls, "get_page_from_preview_token"):
+                """we assume that get_page_from_preview_token validates the token"""
+                return cls.get_page_from_preview_token(token)
+
+    """ 
+    we get parent_id and page_type when previewing a page that has not been
```
zerolab, line 209: parent_id is added for convenience. if I recall correctly we needed to vary things a bit when generating the token. When you have an ID that gives enough to go on.

zerolab (changes requested):
@dopry do you still need help with generating a token for new pages?

https://github.com/torchbox/wagtail-headless-preview/blob/main/src/wagtail_headless_preview/models.py#L75-L80

unpublished pages are no more special than others. You can just do `token = PageModelWithPreviewMixin.create_page_preview().token`

dopry:
I'll try to get another pass on this by EOM.

grapple/types/pages.py
dopry: So this is really just a check to see if headless preview is installed and enabled?

grapple/types/pages.py
dopry: So this is really just a check to see if headless preview is installed and enabled?

grapple/types/pages.py
dopry: we could eliminate the if id: code path here if the page_type was always sent. It might not be as 'fast', but it would reduce the amount of code we need to maintain.

grapple/types/pages.py
dopry: I just wrapped the whole thing in a try/except.

grapple/types/pages.py
dopry: would we be better served with an isHeadlessPreviewEnabled() function?

dopry:
@zerolab I got the tests setup. I'm not happy with how I'm mocking the unsaved page as per my haiku, but it seems to work.  I feel like wagtail-headless-preview is actually a better place for much of this code.  I'll try to make a PR against it in the next few weeks to improve the DX here a little and for anyone else building on it.

zerolab:
uh-oh, I missed this one for 0.19. Will review on Friday and make 0.19.1 release 🙈

dopry:
I was just about to ask. :) I'd love to see previews fixed. At some point in the future I'll try to take a pass on wagtail-headless-preview, but I'd rather those concerns not block this now.

zerolab (approved):

zerolab:
@dopry https://github.com/torchbox/wagtail-grapple/releases/tag/v0.19.1

dopry:
🎉 hooray :tada:

zerolab:
Thank you for your PRs, much appreciated!

Issue #227, opened 2022-04-12
Only first middleware is ran through decorator


seb-b:
If multiple middleware are added to the `register_query_field`, only the first one is ran e.g.

```python

class FistMiddleware(object):
   def resolve(self, next, root, info, **args):
        print("Running first")
        return next(root, info, **args)
        
class SecondMiddleware(object):
   def resolve(self, next, root, info, **args):
        print("Running second")
        return next(root, info, **args)
        
 @register_query_field("demo", middleware=[FistMiddleware, SecondMiddleware])
 class Demo(models.Model):
      ....
```

Then the output is only 'Running first'. Switching the order means only second is outputted.

I think it's because of this line [here](https://github.com/GrappleGQL/wagtail-grapple/blob/main/grapple/middleware.py#L51) where only the first `field_middleware` is returned in the loop instead of yielding all of them.

zerolab:
@seb-b that is a great catch. Do you have time to submit a PR?

PR #281, opened 2022-12-06
Use all middleware defined in decorator

seb-b:
One thing what isn't super clear to me is how the `next` is handled? I'm not sure if this calls it for every iteration of the middleware loop or not until the end.

One thing that this also fixed was function style middleware never worked, it was being yielded and then the exception below it was also triggered


seb-b:
I had a look at the graphene source, looks like they use partials to chain middleware so I've done the same, seems to work ok

seb-b:
The motivation to create this PR was to fix issue #227, it's referenced in a commit message. 

if you check the current implementation [here](https://github.com/torchbox/wagtail-grapple/blob/main/grapple/middleware.py#L41), you can see that the middleware is called in a loop, but it returns on the first iteration. What this PR does is build up a chain middleware and then returns the whole chain, that chain is then what is called, and it'll exit on any of the middleware if it needs to

dopry:
The current implementation is broken.  But your fix isn't quite correct either. This middleware isn't meant to be called in loop.  Each middleware is responsible for calling next, a parent loop shouldn't be doing it.  It should be a function chain. 

I'm wondering if we should be doing something more along the lines of what is done in graphql-core

```
def resolve(self, next, root, info: ResolveInfo, **kwargs):
        field_name = info.field_name
        parent_name = info.parent_type.name
        if field_name in self.field_middlewares and parent_name in ROOT_TYPES:
            middlewares = self.field_middlewares[field_name].copy()
            # see: https://github.com/graphql-python/graphql-core/blob/main/src/graphql/execution/middleware.py
            middleware_manager = MiddlewareManager(middlewares)
            # see: https://github.com/graphql-python/graphql-core/blob/b43c820f9e8674eaf24f16e71beef5f613109e33/src/graphql/execution/execute.py#L514
            resolve_func = middleware_manager.get_field_resolver(next)
            # see: https://github.com/graphql-python/graphql-core/blob/b43c820f9e8674eaf24f16e71beef5f613109e33/src/graphql/execution/execute.py#L527
            return resolve_func(root, info, **kwargs)
       else:
            return next(root, info, **kwargs)
```

seb-b:
The chain is built up using [partials](https://docs.python.org/3/library/functools.html#functools.partial), it returns a function reference, nothing is called until the loop is exited.

I popped your implementation in my dev instance and it caused 5 tests to fail, all to do with middleware not resolving properly (authentication ones mainly). If I could get some clear picture of what is wrong with my implementation that would help me greatly in getting this over the line.

dopry:
I don't doubt that my example may raise other issues... It is just pseudo code copied from the other code base. It is completely untested.    I'll look at it more closely when I get a chance later today. Maybe I was just misreading it.

seb-b:
@dopry no worries! Just rebased

PR #285, opened 2022-12-17
feat: django-graphene3

dopry:
- built on graphql core 3
- asyncio subscriptions

rebased from #143 plus some additional fixes


grapple/middleware.py
```diff
@@ -48,6 +36,6 @@ def resolve(self, next, root, info: ResolveInfo, **args):
         parent_name = info.parent_type.name
         if field_name in self.field_middlewares and parent_name in ROOT_TYPES:
             for middleware in self.field_middlewares[field_name]:
-                return middleware(next, root, info, **args)
+                return middleware(self, next, root, info, **args)
```
zerolab, line 39: would be good to add a note as to why we're passing `self` here, perhaps with a link to docs or code

zerolab (changes requested):
So-so close!
Got a number of notes.
The blockers (linting, I know, I know) and the channels integration

requirements.txt
```diff
@@ -1,8 +1,12 @@
...
 wagtail>=2.15,<5.0
 wagtailmedia
-graphql-core>=2.2.1,<3
-graphene-django>=2.7.1, <2.16.0
+## note graphql-core does not follow semver. backwards compatibiity is only guaranteed for patch
+## releases. Major updates from upstream GraphQL.js will be implemented as minor versions, so there
+## can be breaking changes in minor versions.
+## Do we even need a direct dependency on graphql-core? don't we get it via graphene-django?
+graphql-core==3.2.3
+graphene-django>=3.0.0
```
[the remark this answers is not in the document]
dopry: Will do.

grapple/middleware.py
dopry: I stuck that in because I was getting a parameter mismatch. I'm not exactly sure why the change was required and didn't find any docs for the change.  I'll spend some time trying to figure it out, but I'm not really familiar with how the middleware works.

requirements-channels.txt
```diff
@@ -1,5 +1,4 @@
 -r ./requirements.txt
 
 # Channels-specific requirements
-graphql-ws==0.4.4
```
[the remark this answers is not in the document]
dopry: I'll include RX.

dopry:
I've addressed the outstanding issues, I don't know what to do about adding the self arg to middleware. Are we good as is, or should I dig into it deeper.

grapple/middleware.py
dopry: I did a little more digging and https://github.com/graphql-python/graphene/issues/812 is the most related discussion I can find about why self needs to be there... but what I really don't know is what GrappleMiddleware is intended to do. A deeper change may be necessary for it to work propertly.  I'm pretty sure I'm passing the wrong self into resolve there, which could cause issues with some middleware that expect to have something on their self.

dopry:
@zerolab I figured out the self thing, finally. Middleware classes needed to be instantiated.   Let me know if there are any additional changes you would like to see.

grapple/middleware.py
```diff
@@ -1,26 +1,11 @@
...
-    for middleware in middlewares:
-        if inspect.isfunction(middleware):
-            yield middleware
-        if not hasattr(middleware, MIDDLEWARE_RESOLVER_FUNCTION):
-            raise Exception(
-                "Middleware must be either a class or a function. Got: {}.\nYou can read more about middleware here: https://docs.graphene-python.org/en/latest/execution/middleware/".format(
-                    type(middleware)
-                )
-            )
-        yield getattr(middleware(), MIDDLEWARE_RESOLVER_FUNCTION)
```
zerolab, line 21: Now it makes sense. This bit here instantiates the middleware class and calls the resolver, whereas https://github.com/graphql-python/graphql-core/blob/main/src/graphql/execution/middleware.py#L67 doesn't

will refresh my memory how we pass on the middleware down the chain and see if there's something we can do without folks outright instantiating the middleware classes

zerolab: @dopry ok, I needed more of a break than I thought I did :)

got a bit of time to dig into the code a bit more. we can solve this in https://github.com/torchbox/wagtail-grapple/blob/65a26a9daf4f7a3e937f7eaed90cd8373951be57/grapple/helpers.py#L40-L50 with a `if isinstance(middleware, type)` check.

Will follow this up

zerolab (approved):

zerolab:
Merged in 0f1acd24e981781dfa224d2a17b449647645b165 and parents. Thank you @dopry @fabienheureux

PR #292, opened 2023-01-05
Fix/search ordering

### LINKED ISSUE #290: bug: order is not applied to search results
When using both searchQuery and order, the results are not ordered according to order.

dopry:
closes #290

### DIFF
--- a/grapple/utils.py
+++ b/grapple/utils.py
@@ -103,19 +103,6 @@ def resolve_queryset(
     else:
         qs = qs.all()
 
-    if id is None and search_query:
-        # Check if the queryset is searchable using Wagtail search.
-        if not class_is_indexed(qs.model):
-            raise TypeError("This data type is not searchable by Wagtail.")
-
-        if grapple_settings.ADD_SEARCH_HIT:
-            query = Query.get(search_query)
-            query.add_hit()
-
-        qs = get_search_backend().search(search_query, qs)
-
-        return _sliced_queryset(qs, limit, offset)
-
     if order is not None:
         qs = qs.order_by(*(x.strip() for x in order.split(",")))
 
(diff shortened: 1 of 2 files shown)


dopry:
> The changes LGTM.
> 
> My one nitpick and preference would be to have a separate testcase with the factories in `setUpTestData`.

Can do. Give me a few hours and I'll get it updated,

dopry:
@zerolab updated!

grapple/utils.py
```diff
@@ -127,6 +114,17 @@ def resolve_queryset(
...
+    if id is None and search_query:
+        # Check if the queryset is searchable using Wagtail search.
+        if not class_is_indexed(qs.model):
+            raise TypeError("This data type is not searchable by Wagtail.")
+
+        if grapple_settings.ADD_SEARCH_HIT:
+            query = Query.get(search_query)
+            query.add_hit()
+
+        qs = get_search_backend().search(search_query, qs)
```
zerolab, line 126: So... I about to hit merge, then had my doubts (typical!)
The current version will order by relevance by default as that is the core behaviour. The PR change the logic to only use a custom order_by if that is asked for, which makes perfect sense. In fact the postgres backend does as similar thing, and more - https://github.com/wagtail/wagtail/blob/959a802c76ea760a5893e27783815e580a89d2c4/wagtail/search/backends/database/postgres/postgres.py#L586-L589 🚀

dopry: What was your specific concern? If an explicit order by is provided it should be applied over relevance. Did we lose support for ordering by relevance or is it that the default behavior is changing?

zerolab: all merged and in 0.19 - https://github.com/torchbox/wagtail-grapple/releases/tag/v0.19.0

dopry: :tada:

PR #294, opened 2023-01-09
Update upper bound for Django

zerolab:
graphene-django 3 supports Django >= 3.2


zerolab:
Test failure caused by https://github.com/wagtail/wagtail/issues/9788 with Django 4.1

PR #299, opened 2023-01-14
fix: ordered searches

dopry:


dopry:
@zerolab this should finally get the ordering working properly with search.  The testing would be better if we could use postgres, since sqlite doesn't support search scoring.

zerolab:
For postgres tests: would need something like https://github.com/torchbox/wagtailmedia/blob/main/.github/workflows/test.yml#L28-L35 and probably just the latest wagtail version for testing (because why not?)
and use dj-database-url in settings or something like https://github.com/torchbox/wagtailmedia/blob/main/tests/settings.py#L11-L22, defaulting to sqlite

dopry:
I can totally setup the workflows I have it on plenty of projects. I'm more concerned about how to test with both sqllite and postgres since they'll both return different results or natural search.

dopry:
This currently passes with sqllite and is working in our development and uat environments. (The previous code I primarily unit tested and assumed it was good... bad assumption)

zerolab:
ah, I see now. I wonder if we add another test which we skip unless the db is postgres. One for a follow up PR

zerolab (approved):

dopry:
👍 Thanks for getting this in! I can get off my fork and back up upstream.  🎉

zerolab:
https://github.com/torchbox/wagtail-grapple/releases/tag/v0.19.2 - should be on PyPI on 5 or so mins

Issue #311, opened 2023-02-08
Critical Typo in registering custom Rendition model


engAmirEng:
https://github.com/torchbox/wagtail-grapple/blob/2e7cb3e23f81c3c65e1fddc811aeaed99cd7743c/grapple/actions.py#L135-L142

Line 140 should call **register_image_rendition_model()** not register_image_model()

Fixing this typo will **breake** the server duo to "python-BaseException KeyError: '\_\_module\_\_'" so this issue actually references two issues

And a note for contributors: I Love and appreciate what you have done here in this project 👌❤

engAmirEng:
> Oh, wow. that has been there for some time (#16) How did you stumble upon it?

Yea, I'm doing quite an uncommon project and I discover these kind of stuff in wagtail itself too,  
About this particular one, I am providing the option for users to register their own object storage to be used for pictures, documents and stuff and that's where I needed custom rendition model

jams2:
It looks like this was fixed in #337

PR #312, opened 2023-02-10
Fix problem in registering and querying custom rendition model

engAmirEng:
This feature implemented for (#16) and wasn't working up to now and the issue (#311)
is refrencing this.
As I was exploring this I found some serious bugs and architectural Issues that I'm going to open later.

### DIFF
--- a/grapple/actions.py
+++ b/grapple/actions.py
@@ -22,7 +22,7 @@ from .helpers import field_middlewares, streamfield_types
 from .registry import registry
 from .settings import grapple_settings
 from .types.documents import DocumentObjectType
-from .types.images import ImageObjectType
+from .types.images import ImageObjectType, ImageRenditionObjectType
 from .types.pages import Page, PageInterface
 from .types.rich_text import RichText as RichTextType
 from .types.streamfield import generate_streamfield_union
@@ -122,7 +122,7 @@ def register_model(cls: type, type_prefix: str):
         elif issubclass(cls, AbstractImage):
             register_image_model(cls, type_prefix)
         elif issubclass(cls, AbstractRendition):
-            register_image_model(cls, type_prefix)
+            register_image_rendition_model(cls, type_prefix)
         elif has_wagtail_media and issubclass(cls, AbstractMedia):
             register_media_model(cls, type_prefix)
         elif issubclass(cls, (BaseSiteSetting, BaseGenericSetting)):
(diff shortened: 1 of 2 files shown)


engAmirEng:
@zerolab Do you know about this CI failure?

zerolab:
It is Wagtail 4.2 issue. Will sort it out.
In the meantime, any chance you can add a test?

dopry:
@engAmirEng It's important we try to get tests in to avoid regressions. Tests might have prevented this bug in the first place.  Admittedly the test in grapple are not well organized at the moment and are fully dependent on the example app. 

I think there are two things we would want to test.

1. Is there an image rendition type in the schema?

   we should add a test_graphql_schema.py  in example/home/test. From on of my projects I have https://gist.github.com/dopry/37c6bee35eec0d28483ad8a29fa793bf that could serve as a template. 
  
2. Can we still query renditions?

    I think this could be added in https://github.com/torchbox/wagtail-grapple/blob/main/example/home/test/test_blog.py  in test_blog_body_imagechooserblock and  test_blog_body_imagechooserblock_in_streamblock to ensure renditions can be queried in both contexts. 
   

I'm not sure if testing should hold up this particular PR given it doesn't already have test coverage and we're fixing a pretty obvious production bug. It would be great if you could take a stab at the test.  Ultimately whether tests are required is @zerolab's call.

engAmirEng:
I'll take care of that, might take some time

dopry:
@engAmirEng can you rebase your branch, I want to see if the pre-commit tests are still failing.

engAmirEng:
> @engAmirEng It's awesome the existing tests are passing. The last thing blocking this getting in is some tests. Do you have time to work on those? I'd love to get this in along with the hostname/multitenancy work you've been doing.

Honestly writing tests are not my strong suit, I tried to write the test you are asking for but I couldn't figure it out

dopry:
I understand it can be difficult to get used to writing tests... Normally the approach I take is to write  a test with the old code to demonstrate that the thing I am fixing is broken.  Then I develop my code against that test. Would you be open to some guidance on the test development, and/or another contributor submitting tests to your branch?

engAmirEng:
> I understand it can be difficult to get used to writing tests... Normally the approach I take is to write a test with the old code to demonstrate that the thing I am fixing is broken. Then I develop my code against that test. Would you be open to some guidance on the test development, and/or another contributor submitting tests to your branch?

I will give it another shot today and give you the feedback

zerolab:
I've folded this into #337
Thank you @engAmirEng

engAmirEng:
> I've folded this into #337
> Thank you @engAmirEng 

Happy to see the repo is going forward

PR #324, opened 2023-05-12
chore: re-organize test app

dopry:
- move example to test/testapp
- consolidate tests in one location


dopry:
@zerolab  I'd love to get this in pronto so we're not blocking other PRs that may be implementing tests. I know we've requested tests on a few open PRs.

requirements.txt
```diff
@@ -1,10 +1,10 @@
...
 wagtail>=2.15,<5.0
-wagtailmedia>=0.13
 graphene-django>=3.0.0
 factory-boy==3.2.1
 wagtail-factories>=4.0.0,<4.1.0
 django-cors-headers==3.4.0
+wagtailmedia>=0.13
 wagtail-headless-preview
 # simplified database configuration to support testing multiple DB engines in CI.
 dj-database-url==1.2.0
```
zerolab, line 10: this and `requirements-channels.txt` are test requirements. I suggest moving them to `tests/`

tests/test_grapple.py
```diff
@@ -104,10 +104,10 @@ def test_pages(self):
         self.assertEqual(type(executed["data"]["pages"][0]), dict)
 
         pages_data = executed["data"]["pages"]
-        self.assertEqual(pages_data[0]["contentType"], "home.HomePage")
-        self.assertEqual(pages_data[0]["pageType"], "HomePage")
-        self.assertEqual(pages_data[1]["contentType"], "home.BlogPage")
-        self.assertEqual(pages_data[1]["pageType"], "BlogPage")
+        self.assertEquals(pages_data[0]["contentType"], "testapp.HomePage")
```
zerolab, line 107: as far as I recall `assertEquals` was deprecated in favour of `assertEqual` 
edit: found it https://docs.python.org/3/library/unittest.html#deprecated-aliases

zerolab (changes requested):

zerolab:
@dopry did a quick first pass and left a couple of notes

requirements.txt
dopry: There are some dependencies for the testapp, but we do import from wagtail and django in grapple/*, should I keep those in the /requirements[-channels].txt or just move everything to the testapp? I was kind of thinking to defer altering requirements to the flit pr to trim some scope here since I'm not testing packaging I didn't want to have any unintended impacts.

static/.gitkeep
```diff
@@ -0,0 +1,3 @@
+need to prevent
+WARNINGS:
+?: (staticfiles.W004) The directory '/home/runner/work/wagtail-grapple/wagtail-grapple/static' in the STATICFILES_DIRS setting does not exist.
```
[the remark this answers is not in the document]
dopry: moved.

tests/test_grapple.py
dopry: good catch, updated.

requirements.txt
zerolab: The requirements*.txt files are only used for tests. well, outside of this PR, in the `example` app

The imports under grapple/* should be all handled via the install requirements (i.e. https://github.com/torchbox/wagtail-grapple/blob/main/setup.cfg#L42-L53)

dopry: done.

dopry:
@zerolab I think I've addressed all the issues. This should be ready to review again.

.gitignore
```diff
@@ -139,3 +138,7 @@ Temporary Items
 *.sublime-workspace
 .vscode
 .idea
+
+
+# wagtail
+static/
```
zerolab, line 144:
```suggestion
tests/static/
```

zerolab (approved):

zerolab:
Let's get this in.

PR #329, opened 2023-05-15
Develop/exclude excludeds

engAmirEng:
The new `DefaultField` added, if user wants to only expose some of the fields that are implemented by grapple instead of all of them then they should wrap them in this class, otherwise everything is going to work with no breaking changes and all of the grapple's fields are exposed(as they were before)

### DIFF
--- a/grapple/actions.py
+++ b/grapple/actions.py
@@ -1,7 +1,7 @@
 import inspect
 from collections.abc import Iterable
 from types import MethodType
-from typing import Any, Dict, Type, Union, Callable, Tuple
+from typing import Any, Callable, Dict, Tuple, Type, Union
 
 import graphene
 from django.apps import apps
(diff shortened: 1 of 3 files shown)


engAmirEng:
Just to make sure you got my approach,
I introduced a new class (DefaultField), if user wants to include grapple implemented fields selectively, then wraps them inside this class, if user wants to include all grapple's implemented fields then it's all ok with no breaking changes

zerolab:
Ah, thank you. I did indeed miss that when I had a cursory look. In the future, could you please:

1. include this type of detail in the PR description
2. add corresponding code comments. (aside: I know the code is not consistently commented, but we're chipping away at that)

Now, if we commit to going down this route, we need to follow a 2 release deprecation process similar to how core or Django does it (ref: https://docs.wagtail.org/en/stable/contributing/release_process.html#deprecation-policy) as this would result in backwards incompatible changes.

I have more time on Friday and will look more in detail.

engAmirEng:
@zerolab you are right, I should've describe the details in the PR
I was excited to see if the general approach is considerable

dopry:
@engAmirEng I don't really follow the changes here. I haven't dove deep into the automatic schema generation. I'm not sure what a DefaultField is or why I would want to use it. I don't understand the use case for this change.  Could you update the OP to better explain your goals with these changes and how the implementation works?

engAmirEng:
@dopry Look at #330 and you'll see that I had to skip a test, try to understand why the test is failing

PR #330, opened 2023-05-15
fix resolve_rendition, instantiating ImageRenditionObjectType by hand?!

### LINKED ISSUE #318: querying for "custom_rendition_property" in the example, results in error
"Cannot return null for non-nullable field CustomImageRendition.customRenditionProperty." is the result of querying the "**custom_rendition_property**" while we have this:
https://github.com/torchbox/wagtail-grapple/blob/1bd7972a2c683a610c90be6c7d05dd067ae2eb35/example/images/models.py#L24-L28

engAmirEng:
this fixes #318

not sure why they did instantiate the ImageRenditionObjectType by hand instead of returning the Rendition instance in the first place

the faling test is "test_query_rendition_url_field" and there is no way to fix it until #329 is merged

### DIFF
--- a/grapple/types/images.py
+++ b/grapple/types/images.py
@@ -110,20 +110,9 @@ class ImageObjectType(DjangoObjectType, BaseImageObjectType):
         # Only allowed the defined filters (thus renditions)
         if rendition_allowed(filters):
             try:
-                img = self.get_rendition(filters)
+                return self.get_rendition(filters)
             except SourceImageIOError:
                 return
-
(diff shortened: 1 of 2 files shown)


zerolab:
Thank you for this, @engAmirEng 
Closing in favour of #337 which takes care of test as well

engAmirEng:
> Thank you for this, @engAmirEng 
> Closing in favour of #337 which takes care of test as well

Happy to see the repo is going forward🥳

Issue #332, opened 2023-05-30
Support passing `deprecation_reason` and `description` Graphene type arguments.


Morsey187:
**Description**
Graphene supports the Scalar type argument [deprecation_reason and description](https://docs.graphene-python.org/en/latest/types/scalars/#scalars), this can be useful for maintaining a public endpoint as it provides metadata for tools using schema.json such as GraphiQL.

Theres currently some overlap with the arguments passed from wagtail_grapple's field types to graphenes, i.e `name` and `required`, however, `deprecation_reason` and `description` aren't supported or possible to pass in as keyword arguments, mainly due to how the [GraphQLField constructor](https://github.com/torchbox/wagtail-grapple/blob/b4e256c97548ab742aec8eec908ca3ed5f7b3f48/grapple/models.py#L13) and subsequent utility methods 
 ([get_field_type()](https://github.com/torchbox/wagtail-grapple/blob/b4e256c97548ab742aec8eec908ca3ed5f7b3f48/grapple/actions.py#L157)) pass in arguments and constuct Graphene types.

**Proposal**
I'd like to propose either supporting passing in all arguments as kwargs to graphene, popping wagtail_grapple specific arguments in the constructor for example something similar to the below:

```python
# models.py
class GraphQLField:
    field_name: str
    field_type: str
    field_source: str

    def __init__(
        self, field_name: str, field_type: type = None, required: bool = None, **kwargs
    ):
        # Initiate and get specific field info.
        self.field_name = field_name
        self.field_type = field_type
        self.field_source = kwargs.pop("source", field_name)

        # # Add support for NonNull/required fields
        # if required:
        #     self.field_type = graphene.NonNull(field_type)

        # Legacy collection API (Allow lists):
        self.extract_key = kwargs.pop("key", None)
        is_list = kwargs.pop("is_list", False)
        if is_list:
            self.field_type = graphene.List(field_type)

        self.field_kwargs = kwargs

---
# actions.py
def get_field_type(field):
    # If a tuple is returned then obj[1] wraps obj[0]
    field_wrapper = None
    if isinstance(field, tuple):
        field, field_wrapper = field
        if callable(field):
            field = field()
    print("field.field_kwargs", field.field_kwargs)

    field_type = field.field_type

    if field_type is not None:
        if field_wrapper:
            return field, field_wrapper(field_type, **field.field_kwargs)
        else:
            return field, graphene.Field(field_type, **field.field_kwargs)
```
[...]

PR #333, opened 2023-06-13
Support graphene type arguments

Morsey187:
Closes https://github.com/torchbox/wagtail-grapple/issues/332

Adds support for Scalar type arguments [deprecation_reason and description](https://docs.graphene-python.org/en/latest/types/scalars/#scalars), 

Viewable within GraphiQL (example graphiql query [http://localhost:8000/graphiql/?query=...](http://localhost:8000/graphiql/?query=query%20%7B%0A%20%20settings%20%7B%0A%20%20%20%20%20%20...%20on%20GlobalSocialMediaSettings%20%7B%0A%20%20%20%20%20%20%20%20instagram%0A%20%20%20%20%20%20%20%20youtube%0A%20%20%20%20%20%20%7D%0A%20%20%7D%0A%20%20pages%20%7B%0A%20%20%20%20...on%20BlogPage%20%7B%0A%20%20%20%09%09extraSummary%0A%20%20%09%7D%0A%20%20%7D%0A%7D%0A)) or by running `python manage.py graphql_schema` and viewing the generated schema.json file.

<details><summary>Screenshot</summary>
Screenshot 2023-06-13 at 12 10 40
</details>

**Notes:**
I've tested this manually by updating arguments to graphql_fields in the test app, however, I'm not sure of the best way to  write tests for this feature, I've pushed a file `test_models_types.py` although I'm not sure it's tests generate much value.
I can either: 
- Write different tests to the ones in `test_models_types.py` (with some guidance).
- Finish writing tests similar to the ones in `test_models_types.py`.
- Expand the test app with lots of new fields and models with deprecation_reason set, and add those new fields to the exisitng tests.


grapple/actions.py
```diff
@@ -452,6 +456,8 @@ class Meta:
             ]
         else:
             interfaces = (interface,) if interface is not None else ()
+        # Add description to type if streamfield has a docstring
+        description = cls.__doc__
```
Morsey187, line 460: It should be possible to also depreciate streamfield blocks, I looked into expanding the `register_streamfield_block` decorator to support passing in an additonal argument however, I couldn't seem to find a way to make this work in `build_streamfield_type`. 
As a compromise I thought about using the blocks docstring instead, but happy to remove it if not ideal?

tests/test_models_types.py
```diff
@@ -0,0 +1,195 @@
...
+            self.field_name,
+            "testapp.CustomImage",
+            is_paginated_queryset=True,
+            description=self.description,
+            deprecation_reason=self.deprecation_reason,
+        )()
+        # Use get_field_type to replicate behavior of grapple.actions.load_type_fields
+        field, field_wrapper = get_field_type(field)
+        # Assert field_wrapper type is Field and issubclass of BasePaginatedType
+        assert isinstance(field_wrapper, graphene.Field)
```
zerolab, line 177: we're using unnittest, so `self.AssertIsInstance` etc

grapple/actions.py
zerolab: @Morsey187 if you can update the comment with the extra context (and nitpick "streamfield" -> "the block")

zerolab (changes requested):

grapple/actions.py
jams2: What if, instead of implicitly including a block's docstring as the description, we check the block's metaclass for the presence of a `graphql_description` field? Developers may not want their internal documentation exposed.

tests/test_models_types.py
jams2: Fixed in https://github.com/torchbox/wagtail-grapple/pull/333/commits/c6d8a8bed8ccbb79b0e673713e67d80ec1a98712

grapple/models.py
```diff
@@ -17,16 +17,26 @@ def __init__(
         self.field_name = field_name
         self.field_type = field_type
         self.field_source = kwargs.get("source", field_name)
+        self.description = kwargs.get("description", None)
+        self.deprecation_reason = kwargs.get("deprecation_reason", None)
 
         # Add support for NonNull/required fields
         if required:
-            self.field_type = graphene.NonNull(field_type)
+            assert (
```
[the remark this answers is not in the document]
jams2: Fixed in https://github.com/torchbox/wagtail-grapple/pull/333/commits/c6d8a8bed8ccbb79b0e673713e67d80ec1a98712

grapple/actions.py
jams2: Implemented in https://github.com/torchbox/wagtail-grapple/pull/333/commits/4f006511bc70395df850093226563b6a9302221f

grapple/models.py
```diff
@@ -7,7 +7,7 @@
 # Classes used to define what the Django field should look like in the GQL type
 class GraphQLField:
     field_name: str
-    field_type: str
+    field_type: type
     field_source: str
```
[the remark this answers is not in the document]
jams2: Changed in https://github.com/torchbox/wagtail-grapple/pull/333/commits/4f006511bc70395df850093226563b6a9302221f

grapple/actions.py
jams2: I added docs under `register_streamfield_block`. Not sure what to add to the comment in code from Ben's comment above, or if it's still necessary?

zerolab: I think things are good as they are now. Thanks both

zerolab (approved):

PR #334, opened 2023-06-16
Switch to using Ruff

zerolab:
replaces isort, flake8


dopry (approved):
LGTM and the bare exception situation seems fine.

Issue #340, opened 2023-07-03
Inconsistent error handling in the site query


estyxx:
I've noticed an inconsistency in how wagtail-grapple handles errors for the `site` query, which takes `id` and `hostname` as parameters.

When an incorrect `id` is provided, the query appropriately returns `null`, meaning the requested site does not exist. However, when a non-existent `hostname` is provided, it raises an unhandled `DoesNotExist` exception. This exception subsequently results in a `GraphQLError` that is returned in the "errors" response.

The current implementation adds complexity to differentiating between real unhandled errors and expected behaviours, as it behaves differently based on the input parameter.

From a conceptual standpoint, it seems incorrect to raise an unhandled error when the site does not exist, as it is an expected scenario that can occur in normal operation. 

It would be beneficial for the `site` query to handle these errors consistently across both `id` and `hostname` parameters. This would make error handling more predictable and user-friendly.

zerolab:
Fixed by #342

PR #342, opened 2023-07-05
Consolidate handling when resolving Sites and avoid making assumptions about what data to return when it's ambiguous

kbayliss:
https://github.com/torchbox/wagtail-grapple/issues/340 identifies that the handling of resolving sites is inconsistent depending on the context.

This MR addresses that so that:

1. A `GraphQLError` is no longer raised if querying for a `Site` with a hostname that doesn't exist (as this is an expected error) - similar to what already happens when querying for a `Site` by ID that doesn't exist
2. Consolidates the handling of site resolving to reduce code
3. We no longer return `Page`s for the first `Site` when querying for `Page`s with an erroneous `Site` filter (i.e. providing a `hostname=does.not.exist` query filter should not return `Page`s for the first `Site` - it should return `None`).
4. We no longer return `SiteSetting`s for the first `Site` when querying for `SiteSetting`s without a site filter when there are multiple `Site`s.

Also, I had to remove a migration dependency for `taggit` when initialising an empty database - I didn't look into why, though I did confirm that `taggit` migrations are still run as Wagtail requires 'em.

### DIFF
--- a/grapple/types/settings.py
+++ b/grapple/types/settings.py
@@ -1,5 +1,4 @@
 import graphene
-from graphql.error import GraphQLError
 from wagtail.contrib.settings.models import BaseSiteSetting
 from wagtail.models import Site
 
@@ -34,14 +33,7 @@ def SettingsQuery():
                 site_hostname = kwargs.pop("site", None)
 
                 if site_hostname is not None:
-                    try:
-                        site = resolve_site(site_hostname)
-                    except Site.MultipleObjectsReturned as err:
-                        raise GraphQLError(
-                            "Your 'site' filter value of '{}' returned multiple sites. Try adding a port number (for example: '{}:80').".format(
-                                site_hostname, site_hostname
-                            )
-                        ) from err
+                    site = resolve_site(hostname=site_hostname)
                 else:
                     site = None
 
@@ -53,8 +45,13 @@ def SettingsQuery():
 
                     if site and issubclass(setting._meta.model, BaseSiteSetting):
                         return setting._meta.model.objects.filter(site=site).first()
-                    else:
+
+                    # If there's only one Site, we can reliably return the
+                    # correct SiteSetting here.
+                    if Site.objects.all().count() == 1:
                         return setting._meta.model.objects.first()
+                    else:
+                        return None
 
             # Return all settings.
             def resolve_settings(self, info, **kwargs):
@@ -63,14 +60,7 @@ def SettingsQuery():
                 site_hostname = kwargs.pop("site", None)
 
                 if site_hostname is not None:
-                    try:
-                        site = resolve_site(site_hostname)
-                    except Site.MultipleObjectsReturned as err:
-                        raise GraphQLError(
-                            "Your 'site' filter value of '{}' returned multiple sites. Try adding a port number (for example: '{}:80').".format(
-                                site_hostname, site_hostname
-                            )
-                        ) from err
+                    site = resolve_site(hostname=site_hostname)
                 else:
                     site = None
 
--- a/grapple/types/settings.py
+++ b/grapple/types/settings.py
@@ -4,7 +4,7 @@ from wagtail.contrib.settings.models import BaseGenericSetting, BaseSiteSetting
 from wagtail.models import Site
 
 from ..registry import registry
-from ..utils import resolve_site
+from ..utils import resolve_site_by_hostname
 
 
 def SettingsQuery():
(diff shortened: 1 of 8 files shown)


grapple/types/settings.py
```diff
@@ -53,8 +45,13 @@ def resolve_setting(self, info, **kwargs):
...
                     if site and issubclass(setting._meta.model, BaseSiteSetting):
                         return setting._meta.model.objects.filter(site=site).first()
-                    else:
+
+                    # If there's only one Site, we can reliably return the
+                    # correct SiteSetting here.
+                    if Site.objects.all().count() == 1:
                         return setting._meta.model.objects.first()
+                    else:
+                        return None
```
kbayliss, line 54: Hmm, wonder if this should return a GraphQLError?

tests/testapp/migrations/0001_initial.py
```diff
@@ -28,7 +28,6 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ("wagtailcore", "0078_referenceindex"),
-        ("taggit", "0005_auto_20220424_2025"),
```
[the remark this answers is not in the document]
kbayliss: From the PR description:

> Also, I had to remove a migration dependency for `taggit` when initialising an empty database - I didn't look into why, though I did confirm that `taggit` migrations are still run as Wagtail requires 'em.

grapple/types/sites.py
```diff
@@ -70,24 +72,21 @@ class Mixin:
...
-                    return resolve_site(hostname)
-                except Site.MultipleObjectsReturned as err:
-                    raise GraphQLError(
-                        "Your 'hostname' filter value of '{}' returned multiple sites. Try adding a port number (for example: '{}:80').".format(
-                            hostname, hostname
-                        )
-                    ) from err
+        def resolve_site(self, info, **kwargs) -> Optional[Site]:
+            """
+            Return a single `Site` object, identified by ID or hostname.
```
estyxx, line 84:
```suggestion
            Return a single `Site` object, identified by ID or hostname.
            If neither ID nor hostname is provided, the function will return None.
```
it will implicitly return None, and there is the type annotation but since there is no explicit `return None` maybe is worth to mention in docstring? 
Or it's too much?

tests/test_grapple.py
```diff
@@ -1423,6 +1463,10 @@ def setUp(self):
         )
 
     def test_query_single_setting(self):
+        # This only works if there is a single Site, so let's make that true.
```
estyxx, line 1466: I don't get this... Why are we restricted to one setting? Shouldn't we be able to retrieve the desired setting regardless of how many are in the database?

tests/testapp/migrations/0001_initial.py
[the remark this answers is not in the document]
estyxx: Ah sorry I missed that :)

grapple/types/sites.py
kbayliss: Good suggestion - done

estyxx (approved):
LGTM! Nice refactoring 😍

grapple/utils.py
```diff
@@ -9,39 +12,68 @@
 from .types.structures import BasePaginatedType, PaginationType
 
 
-def resolve_site(hostname):
+def resolve_site(
+    *,
+    id: Optional[int] = None,
+    hostname: Optional[str] = None,
+    hostname_filter_name: Optional[Literal["site", "hostname"]] = None,
+) -> Optional[Site]:
```
zerolab, line 20: I have to say I am not enamored with this. the if dance just doesn't sit well with me.
My preference here would be to split this into `resolve_site_by_id(id: int)` and `resolve_site_by_hostname(hostname: str, filter_name: Literal["site", "hostname"])` and then the logic inside is easier to follow, and we don't need to mix .get/DoesNotExist with MultipleObjectsReturned

zerolab (changes requested):
My only gripe is with the `resolve_site` util. The rest LGTM.
Will ask for a changelog and release notes blurb too ;)

grapple/utils.py
kbayliss: Good shout - because the method originally handled both I didn't consider changing it, though this is much simpler!

grapple/utils.py
```diff
@@ -12,56 +12,49 @@
...
+        # This is an expected error, so should not raise a GraphQLError.
+        return None
 
-    If resolving via hostname and two `Site` records exist with the same
-    hostname, you must include the port to disambiguate.
+
+def resolve_site_by_hostname(
+    *,
+    hostname: str,
+    hostname_filter_name: Literal["site", "hostname"],
```
zerolab, line 33: nitpick:
```suggestion
    filter_name: Literal["site", "hostname"],
```

;)

kbayliss: Done!

zerolab (approved):

PR #350, opened 2023-08-02
Uniform ID Type for pages and page Queries

estyxx:
Currently, in the Wagtail GraphQL API of wagtail-grapple, there is an inconsistency in the input types for the `pages` and `page` queries. While the `pages` query expects an `ID` as an input, the `page` query expects an `Int`. This has been leading to unnecessary conversion between string and integer.

#### Proposed Changes
I propose to make the input types consistent by changing the `id` in the `page` query to be of type `ID` (the same as in the `pages` query).

The change can be made in the following file:
[grapple/types/pages.py#L336](https://github.com/torchbox/wagtail-grapple/blob/a40354200a5977bb792cad73d8f667e297666eae/grapple/types/pages.py#L336)

By changing the return type to `graphene.ID`, it ensures uniformity across the queries and eradicates the need for conversions between string and integer.

I updated the documentation as well.

### DIFF
--- a/docs/general-usage/graphql-types.rst
+++ b/docs/general-usage/graphql-types.rst
@@ -80,7 +80,7 @@ The singular ``page`` field accepts the following arguments:
 
 .. code-block:: graphql
 
-    id: Int                       # Can be used on it's own
+    id: ID                        # Can be used on it's own
     slug: String                  # Can be used on it's own
     urlPath: String               # Can be used on it's own
     token: String                 # Must be used with one of the others. Usually contentType
--- a/grapple/helpers.py
+++ b/grapple/helpers.py
@@ -74,7 +74,7 @@ def register_query_field(
         field_type = lambda: registry.models[cls]  # noqa: E731
         field_query_params = query_params
         if field_query_params is None:
-            field_query_params = {"id": graphene.Int()}
+            field_query_params = {"id": graphene.ID()}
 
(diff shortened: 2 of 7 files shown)


zerolab:
@estyxx the pre-commit failure is from the recent ruff update. can you update [`tests/settings.py:L190`](https://github.com/torchbox/wagtail-grapple/blob/main/tests/settings.py#L190) to be `# noqa: S105`?

estyxx:
Mmm it's already commented with #noqa S105 on my branch 🤔

jams2:
Forgive me if this is an ignorant question, but why should we use `ID` rather than `Int`? I assume that the `id` field maps to `Page.id`, which is an integer column.

zerolab:
The benefit of using `ID` over `Int` everywhere is that it can support models (e.g. snippets) with non-int primary keys. Also https://graphql.com/learn/scalars-objects-lists/#the-id-type mentions lots of frameworks/tools have better support for `ID` out of the box.

Now, this PR only changes it for pages, so if we commit, we should go all the way.

The reason I haven't merged this yet is that I am still deciding as it is a potentially breaking change

zerolab (approved):
@estyxx having "slept on it", it is a worthy change.
Any chance you could rebase?

estyxx:
@zerolab I rebased this PR :)

PR #356, opened 2023-08-25
Pass the `GraphQLResolveInfo` object to the StreamField callables

zerolab:
Passes the `GraphQLResolveInfo` object to the custom StreamField resolvers.
This is more in line with the expected GraphQL resolver definitions. This would allow using the request context (`info.context`) for page url resolving (`page.get_url(request=info.context)`) in a multi-site context

Builds on #220 and #322.

> [!WARNING]
> This is a breaking change and requires existing resolver signatures to be updated.
> ```python
> # before
> def get_field(self, value=None):
>    ...
>
> def get_field(self, *, value=None):
>    ...
>
> # after
> def get_field(self, info, value=None):
>    ...
>
> def get_field(self, *, info, value=None):
>    ...
>
> # or, type hinted
> from graphql import GraphQLResolveInfo
>
> 
> def get_field(self, info: GraphQLResolveInfo, value: dict[str, Any] | None):
>     ...
> ```


tests/testapp/blocks.py
```diff
@@ -160,15 +166,17 @@ def simple_string(self) -> str:
...
         # Should not be used as we define `source="get_simple_string_method"`.
         raise Exception
 
     def get_simple_string_method(
         self,
-        values: Dict[str, Any] = None,
-    ) -> Optional[str]:
+        info: GraphQLResolveInfo,
+        values: dict[str, Any] = None,
+    ) -> str | None:
```
kbayliss, line 179: **Question:** Why `str | None` instead of `Optional[str]`?

zerolab: this is the Python 3.10+ notation (https://docs.python.org/3/library/typing.html#typing.Optional)

grapple/actions.py
```diff
@@ -418,7 +418,7 @@ def custom_cls_resolver(*, cls, graphql_field):
         elif callable(getattr(cls, graphql_field.field_source)):
             return lambda self, instance, info, **kwargs: getattr(
                 klass, graphql_field.field_source
-            )(values=get_all_field_values(instance=instance, cls=cls), **kwargs)
+            )(info, values=get_all_field_values(instance=instance, cls=cls), **kwargs)
```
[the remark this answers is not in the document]
zerolab: I went for consistency. Graphene resolvers pass `info` as positional args - https://docs.graphene-python.org/en/latest/types/objecttypes/#graphql-execution-info-info

tests/testapp/blocks.py
```diff
@@ -160,15 +166,17 @@ def simple_string(self) -> str:
 
     def simple_string_method(
         self,
-        values: Dict[str, Any] = None,
+        info: GraphQLResolveInfo,
```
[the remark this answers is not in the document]
zerolab: Given #350 adds breaking changes, I thought we'd rip the bandaid (and maintain consistency with the other resolvers)

tests/testapp/blocks.py
kbayliss: Ja, though is there a reason to choose a union over Optional? I find the union harder to read than `Optional`, though that's just a preference.

grapple/actions.py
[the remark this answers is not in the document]
kbayliss: As we already have a `values` kwarg and our `info` position differs from Graphene's, i'd choose to use a keyword here. Unless it's not possible, choosing keyword arguments is always the way imo. It also makes using it feel slightly nicer.

This:

```
def my_method(self, *, info: GraphQLResolveInfo, values: dict[str, Any])
```

Feels more typical than:

```
def my_method(self, info: GraphQLResolveInfo, *, values: dict[str, Any])
```

tests/testapp/blocks.py
zerolab: less stuff to type. not fussed about it, really. will change

zerolab: ah, can't do that as we're using pyupgrade which enforces pep 604 typing rewrites - https://github.com/asottile/pyupgrade?tab=readme-ov-file#pep-604-typing-rewrites

grapple/actions.py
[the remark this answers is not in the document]
zerolab: updated

kbayliss (approved):
LGTM!

kbayliss:
Suggestion: It would be good to update the `testapp` with a method that uses the `info` - perhaps to fetch the page for request?

PR #359, opened 2023-08-30
Add SVG support

### LINKED ISSUE #345: Add equivalent of wagtailimages_tags.image's "preserve-svg" argument
When using Wagtail's `image` template tag, developers can provide a `preserve-svg` argument ([docs](https://docs.wagtail.org/en/stable/topics/images.html#svg-images)). When supplied, this argument will cause Wagtail to limit the filters applied to SVGs to only those that would not require rasterisation - i.e. resize and crop operations. If not supplied, attempting to apply a rasterisation-requiring operation to an SVG will result in an error being raised, as Willow cannot perform the operation.

There is no equivalent to this in `grapple`. If acceptable, I would like to add this functionality. It looks like it would be sufficient to add a new argument to the `rendition` field ([here](https://github.com/torchbox/wagtail-grapple/blob/bf5cd104605e112a57ab26b6ee874076e4991cce/grapple/types/images.py#L73)), and update the `resolve_rendition` method with extra conditional logic ([here](https://github.com/torchbox/wagtail-grapple/blob/bf5cd104605e112a57ab26b6ee874076e4991cce/grapple/types/images.py#L92)).

zerolab:
Fixes #345

To-Do:
- [x] output tests
- [x] tests that ensure `to_svg_safe_spec` is run for SVG images


tests/test_image_types.py
```diff
@@ -0,0 +1,337 @@
...
+    def setUpTestData(cls):
+        cls.image_model = get_image_model()
+        cls.example_image = wagtail_factories.ImageFactory(title="Example Image")
+        cls.example_image.full_clean()
+        cls.example_image.save()
+
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+        shutil.rmtree(settings.MEDIA_ROOT, ignore_errors=True)
```
kbayliss, line 25: **Suggestion:** Add a comment to clarify that this media directory only exists within the `tests` dir and we're not removing the project's main `media` directory.

tests/test_image_types.py
```diff
@@ -0,0 +1,337 @@
...
+from grapple.types.images import rendition_allowed
+
+
+class ImageTypesTest(BaseGrappleTestWithIntrospection):
+    @classmethod
+    def setUpTestData(cls):
+        cls.image_model = get_image_model()
+        cls.example_image = wagtail_factories.ImageFactory(title="Example Image")
+        cls.example_image.full_clean()
+        cls.example_image.save()
```
kbayliss, line 20: Do we need to call `full_save()` or `save()` here? I think factories will save the instance/object by default: https://factoryboy.readthedocs.io/en/stable/reference.html#factory.CREATE_STRATEGY

On NCBI, we only call .save() if we've manipulated the object after it's created - and all tests run there.

grapple/types/images.py
```diff
@@ -71,64 +106,75 @@ class ImageObjectType(DjangoObjectType):
...
         Render a custom rendition of the current image.
         """
-        filters = "|".join([f"{key}-{val}" for key, val in kwargs.items()])
+        preserve_svg = kwargs.pop("preserve_svg", False)
+        filter_specs = "|".join([f"{key}-{val}" for key, val in kwargs.items()])
 
         # Only allowed the defined filters (thus renditions)
-        if not rendition_allowed(filters):
+        if not rendition_allowed(filter_specs):
+            return
```
kbayliss, line 130: It would be great to return specific exceptions here (and elsewhere) that can be handled by the client instead of returning nothing.

tests/test_grapple.py
```diff
@@ -70,6 +67,31 @@ def setUp(self):
         executed = self.client.execute(query)
         self.available_queries = executed["data"]["__schema"]["queryType"]["fields"]
 
+    def query_schema_by_type(self, object_type: str):
```
kbayliss, line 70: Add docstring

tests/test_image_types.py
```diff
@@ -0,0 +1,337 @@
...
+        {
+            images {
+                id
+                url
+                src
+            }
+        }
+        """
+
+        executed = self.client.execute(query)
```
kbayliss, line 44: As we're not checking `"images" in executed["data"]` or any other keys, it might be nice to change this to:

```suggestion
        executed = self.client.execute(query)["data"]["images"][0]
```

To make your asserts simpler, e.g.:

```
self.assertEqual(executed["id"], str(self.example_image.id))
```

tests/test_image_types.py
zerolab: good spot. was a remnant from `<insert reason that I don't recall now>`

tests/test_image_types.py
zerolab: reworked this entirely

tests/test_image_types.py
```diff
@@ -0,0 +1,346 @@
...
+            self.assertEqual(mapping["isSvg"]["type"]["kind"], "NON_NULL")
+            self.assertEqual(rendition_args["preserveSvg"]["type"]["name"], "Boolean")
+            self.assertEqual(
+                rendition_args["preserveSvg"]["description"],
+                "Prevents raster image operations (e.g. `format-webp`, `bgcolor`, etc.) being applied to SVGs. "
+                "More info: https://docs.wagtail.org/en/stable/topics/images.html#svg-images",
+            )
+        else:
+            self.assertNotIn("isSvg", mapping)
+            self.assertNotIn("preserveSvg", rendition_args)
```
kbayliss, line 246: **Issue:** A test should only have one expected result.

Can this be placed within `ImageTypesTestWithSVG` with all the other Wagtail5+ SVG tests?

Then we can alter this to:

```suggestion
    @unittest.skipIf(WAGTAIL_VERSION >= (5, 0), "Not applicable to Wagtail versions 5.0+"
    def test_svg_fields_not_in_schema(self):
        """
        Ensure SVG-related fields and arguments are not included within the schema if we're testing against Wagtail < 5.0.
        """
        
        results = self.introspect_schema_by_type(self.image_model.__name__)
        mapping = {
            field["name"]: field for field in results["data"]["__type"]["fields"]
        }
        rendition_args = {arg["name"]: arg for arg in mapping["rendition"]["args"]}
        
        self.assertNotIn("isSvg", mapping)
        self.assertNotIn("preserveSvg", rendition_args)
```

https://docs.python.org/3/library/unittest.html#unittest.skipIf

tests/test_image_types.py
```diff
@@ -0,0 +1,346 @@
...
+
+if WAGTAIL_VERSION >= (5, 0):
+    from wagtail.images.tests.utils import get_test_image_file_svg
+
+    class ImageTypesTestWithSVG(BaseGrappleTest):
+        @classmethod
+        def setUpTestData(cls):
+            super().setUpTestData()
+            cls.example_svg_image = wagtail_factories.ImageFactory(
+                title="Example SVG Image",
```
kbayliss, line 257: **Nitpick:** `Example SVG Image` is not used anywhere so could be omitted here.

tests/test_image_types.py
```diff
@@ -0,0 +1,346 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
+        cls.example_image.file.delete(False)
+
+    def test_properties_on_saved_example_image(self):
+        example_img = self.image_model.objects.first()
+
+        self.assertEqual(example_img.id, self.example_image.id)
+        self.assertEqual(example_img.title, "Example Image")
```
kbayliss, line 32: Is this testing factory boy functionality? Is it useful to do that?

kbayliss (changes requested):
@zerolab this is looking awesome! I love the types/hints - so much easier to read 🙌🏻.

grapple/types/images.py
```diff
@@ -71,64 +109,80 @@ class ImageObjectType(DjangoObjectType):
...
-        except SourceImageIOError:
-            return
+            if not filter_specs:
+                raise TypeError(
+                    "No valid filter specs for SVG. "
+                    "See https://docs.wagtail.org/en/stable/topics/images.html#svg-images for details."
+                )
+
+        # previously we wrapped this in a try/except SourceImageIOError block.
+        # Removed to allow the error to bubble up in the response ("errors") and be handled by the user.
```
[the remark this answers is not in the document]
zerolab: This will go as a note in the changelog and release notes. I do want to keep it for a while as it is easier to have inline code context than scour release notes.

tests/test_image_types.py
zerolab: maybe. no.
and at the same time I was limiting the amount of swooping changes in one PR...

tests/test_image_types.py
zerolab: I prefer to keep it

Issue #231, opened 2022-06-17
Add a way to define custom Interfaces


Didericis:
We're using this project to define some models that inherit from other models, both of which are exposed as graphql types via grapple. It'd be really awesome if there were some way we could declare an interface, and then have grapple pay attention to model inheritance to apply those interfaces. Here's an example using two classes where we were wanting this feature.

```py
@register_streamfield_block
class BaseTier(blocks.StructBlock):
    background_image = ImageChooserBlock(required=False)
    block_style = blocks.ChoiceBlock(
        BLOCK_STYLES,
        default=('white', 'White'),
        required=False
    )
    floating_image = blocks.StreamBlock([
        ('floating_image', ImageBlock())
    ], min_num=0, max_num=1, required=False)

    block_tail = blocks.BooleanBlock(default=False, required=False)

    # maybe there could be a new attribute like this?
    graphql_interface_fields = [
        GraphQLImage('background_image'),
        GraphQLString('block_style'),
        GraphQLStreamfield('floating_image'),
        GraphQLBoolean('block_tail')
    ]

@register_streamfield_block
class HeroTier(BaseTier):
    background_video = VideoChooserBlock(required=False)
    type = blocks.ChoiceBlock(choices=[
        ('default', 'Default'),
        ('big', 'Big'),
        ('event', 'Event'),
        ('profile', 'Profile')
    ], default=('default', 'Default'))
    header = blocks.CharBlock(required=False)
    dek = blocks.CharBlock(required=False)
    cta = blocks.StreamBlock([
        ('cta', CTABlock())
    ], min_num=0, max_num=1, required=False)

    # these would be in addition to the interface fields
    graphql_fields = [
        GraphQLMedia('background_video'),
        GraphQLString('type'),
        GraphQLString('header'),
        GraphQLString('dek'),
        GraphQLStreamfield('cta')
    ]
```

The main reason we're wanting this is that we're currently not able to use fragments on the frontend across tiers. Ideally, we'd like to create a `BaseTier` fragment that could be used on `HeroTier` (and other tiers we have) so that we always get fields like `blockStyle` and `blockTail`.

```gql
fragment CompleteBaseTier on BaseTier {
    id
    blockTail
    blockStyle
}

# we can't currently do this, as HeroTier does not implement a BaseTier interface
fragment CompleteHeroTier on HeroTier {
    ...CompleteBaseTier
    header
    dek
}
```

[...]

PR #365, opened 2023-09-05
SteamField block extra interfaces (part 1)

mgax:
Refs. https://github.com/torchbox/wagtail-grapple/issues/231

Allow specifying extra interfaces when registering a streamfield block:

```py
@register_streamfield_block(extra_interfaces=(CustomInterfaceForBlock,))
class BlockWithCustomInterface(blocks.StructBlock):
    # ...
```

### DIFF
--- a/grapple/helpers.py
+++ b/grapple/helpers.py
@@ -17,22 +17,28 @@ streamfield_types = []
 field_middlewares = {}
 
 
-def register_streamfield_block(cls):
-    base_block = None
-    for block_class in inspect.getmro(cls):
-        if block_class in registry.streamfield_blocks:
-            base_block = registry.streamfield_blocks[block_class]
+def register_streamfield_block(cls=None, extra_interfaces=()):
+    def decorator(cls):
+        base_block = None
+        for block_class in inspect.getmro(cls):
+            if block_class in registry.streamfield_blocks:
+                base_block = registry.streamfield_blocks[block_class]
 
-    streamfield_types.append(
-        {
-            "cls": cls,
-            "type_prefix": "",
-            "interfaces": (StreamFieldInterface,),
-            "base_type": base_block,
-        }
-    )
+        streamfield_types.append(
+            {
+                "cls": cls,
+                "type_prefix": "",
+                "interfaces": (StreamFieldInterface, *extra_interfaces),
+                "base_type": base_block,
+            }
+        )
 
-    return cls
+        return cls
+
+    if isinstance(cls, type) and not extra_interfaces:
+        return decorator(cls)
+
+    return decorator
 
 
 def register_graphql_schema(schema_cls):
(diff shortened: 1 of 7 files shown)


tests/testapp/blocks.py
```diff
@@ -309,3 +326,4 @@ class StreamFieldBlock(blocks.StreamBlock):
     block_with_name = BlockWithName()
     advert = SnippetChooserBlock("testapp.Advert")
     person = SnippetChooserBlock("testapp.Person")
+    block_with_custom_interface = BlockWithCustomInterface()
```
zerolab, line 329: nitpick: the convention is to end with what it is. so something like `CustomInterfaceBlock`. 

and `CustomInterfaceForBlock` -> `CustomBlockInterface`. I had a few moment expanding the GraphQL query locally, and exploring the schema in GraphiQL

grapple/helpers.py
```diff
@@ -17,22 +17,28 @@
...
+        base_block = None
+        for block_class in inspect.getmro(cls):
+            if block_class in registry.streamfield_blocks:
+                base_block = registry.streamfield_blocks[block_class]
+
+        streamfield_types.append(
+            {
+                "cls": cls,
+                "type_prefix": "",
+                "interfaces": (StreamFieldInterface, *extra_interfaces),
```
zerolab, line 31: I am torn between calling `extra_interfaces` just `interfaces`.

Also, at the moment if you pass `(StreamFieldInterface, MyCustomInterface)`, the whole GraphQL endpoint crashes with `Type TheStreamFieldBlock can only implement StreamFieldInterface once."`. Let's be more defensive

zerolab (changes requested):
Thank you @mgax,

I've added a few thoughts/suggestions.

The docs need updating too ([ref](https://github.com/torchbox/wagtail-grapple/blob/683418a76b98a2d7ab397996aa51567739eeb9ff/docs/general-usage/decorators.rst#register_streamfield_block))

grapple/helpers.py
```diff
@@ -17,22 +17,28 @@
...
-            "cls": cls,
-            "type_prefix": "",
-            "interface": StreamFieldInterface,
-            "base_type": base_block,
-        }
-    )
-
-    return cls
+def register_streamfield_block(cls=None, extra_interfaces=()):
+    def decorator(cls):
```
zerolab, line 21: also, let's change `cls` to `klass` as it mirrors `cls` from parent function

grapple/helpers.py
mgax: Renamed to `interfaces=`, it's what I was initially using. And the `interfaces` tuple is now constructed from a set so that `StreamFieldInterface` gets deduplicated.

grapple/helpers.py
mgax: I've changed the name in the outer decorator to keep the inner decorator variable consistent with the dict key.

zerolab:
@mgax merging this. As noted in Slack, I will move to declaring the additional interfaces via a property on the model, so will adapt this

PR #366, opened 2023-09-11
Define additional interfaces via `graphql_interfaces`

zerolab:
This allows defining additional interfaces on your models via the `graphql_interfaces` attribute.

Also does a tiny bit of tidy-up in docs, but not too much. New docs page: https://wagtail-grapple--366.org.readthedocs.build/en/366/general-usage/interfaces.html


Morsey187 (approved):
LGTM! 
I've Reviewed and tested this locally, just a few nickpicks around the documentation mainly

docs/general-usage/graphql-types.rst
```diff
@@ -229,7 +151,7 @@ You can also query a setting by model name:
 
 .. code-block:: graphql
 
-    {
+    query {
```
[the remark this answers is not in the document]
zerolab: I may have missed one or two 🙈  Will address and go with a fine comb

docs/general-usage/interfaces.rst
```diff
@@ -0,0 +1,163 @@
...
+``PageInterface``
+-----------------
+
+The default GraphQL interface for all Wagtail ``Page``-derived models.
+
+This is accessible through the ``pages`` or ``page`` field on the root query type.
+
+The interface exposes the following fields, following the Wagtail Page model fields and properties:
+
+.. code-block:: graphql
```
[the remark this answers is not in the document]
zerolab: we're using https://pypi.org/project/pygments-graphql/ but it looks like that doesn't do any good. Will go forth and remove all `code-block:: graphql` refs

docs/general-usage/interfaces.rst
```diff
@@ -0,0 +1,166 @@
...
+
+    blocks: [StreamFieldInterface!]!  # a list of blocks in the StreamBlock or StructBlock
+
+
+
+Adding your own interfaces
+--------------------------
+
+To add your own interfaces to your model, define the ``graphql_interfaces`` attribute on it. The attribute should be
+a tuple of interfaces to add to the model.
```
mgax, line 114: **Nitpick:** Would this look better as a list? Django historically used tuples for class attributes on models, forms, ModelAdmins, and such, but have been transitioning to lists. [Python docs](https://docs.python.org/3/tutorial/datastructures.html#tuples-and-sequences) also have this to say:

> Tuples are immutable, and usually contain a heterogeneous sequence of elements that are accessed via unpacking (see later in this section) or indexing (or even by attribute in the case of namedtuples). Lists are mutable, and their elements are usually homogeneous and are accessed by iterating over the list.

pyproject.toml
```diff
@@ -47,7 +47,6 @@ docs = [
     "sphinx-wagtail-theme>=6.0,<7.0",
     "sphinxcontrib-spelling>=8.0,<9.0",
     "sphinx_copybutton>=0.5",
-    "pygments-graphql==1.0.0",
```
mgax, line 50: **Question:** Out of curiosity, why doesn't pygments-graphql work? I checked the current version of the docs on RTFD and indeed the queries are not being highlighted.

mgax (commented):
The API to specify interfaces makes more sense than my decorator patch, and I think the code and docs improvements are spot on. I just have a nitpick about lists vs tuples. Can't wait to have this released :)

pyproject.toml
Morsey187: Originally apart of this thread https://github.com/torchbox/wagtail-grapple/pull/366#discussion_r1327305076, I think it's because when it's being used for types, for example:

```
id: ID
limit: PositiveInt
offset: PositiveInt
``` 

It's not technically valid syntax, so results in odd rendering, although I'm wondering if we should remove all `code-block:: graphql` refs as I thought it worked for valid query syntax we make use of, for example:

```
query {
    pages {
        ...on BlogPage {
            the_custom_field
        }
    }
}

```

If not, it's probably best to remove it completely.

zerolab: Had a look at the source code for pygments-graphql and it should have worked for types.
The tl;dr it was something on my list to tweak and play with because it wasn't working. 
May as well drop it - I am all for fewer dependencies

docs/general-usage/interfaces.rst
zerolab: Both will work as we're unpacking with `*`. Can tweak the language to say tuple/list

pyproject.toml
mgax: I think full queries aren't getting rendered properly either, e.g. the example at the end of this section: https://wagtail-grapple.readthedocs.io/en/latest/general-usage/model-types.html#module-0 https://github.com/torchbox/wagtail-grapple/blob/d6d0ab7d2baef8d0e56dbbbed1a656b7adc61db0/docs/general-usage/model-types.rst#L127-L135

zerolab: the whole thing doesn't work, hence dropping it :D

docs/general-usage/interfaces.rst
zerolab: tweaked to suggest list first, tuple second

Issue #369, opened 2023-09-26
Add preserveSvg argument to ImageObjectType.srcSet


mgax:
The `preserveSvg` argument, introduced in https://github.com/torchbox/wagtail-grapple/commit/802428fd0e0e56be484a1f279917739a83cc435e on the `rendition` field, would also be super useful on the `srcSet` field, to avoid such errors when requesting `srcSet(sizes: [360, 1024], format: "webp")` on an SVG image:

> 'SvgImage' object has no attribute 'save_as_webp'

Ref: #359

PR #370, opened 2023-09-26
Implement `preserveSvg` flag for `srcSet()`

mgax:
Fixes https://github.com/torchbox/wagtail-grapple/issues/369

### DIFF
--- a/grapple/types/images.py
+++ b/grapple/types/images.py
@@ -59,6 +59,24 @@ def get_rendition_field_kwargs() -> dict[str, graphene.Scalar]:
     return kwargs
 
 
+def get_src_set_field_kwargs() -> dict[str, graphene.Scalar]:
+    """
+    Returns a list of kwargs for the srcSet field.
+    Extracted for convenience, to accommodate for the conditional logic needed for various Wagtail versions.
+    """
+    kwargs = {
+        "sizes": graphene.List(graphene.Int),
+        "format": graphene.String(),
+    }
+    if WAGTAIL_VERSION > (5, 0):
+        kwargs["preserve_svg"] = graphene.Boolean(
+            description="Prevents raster image operations (e.g. `format-webp`, `bgcolor`, etc.) being applied to SVGs. "
+            "More info: https://docs.wagtail.org/en/stable/topics/images.html#svg-images"
+        )
+
+    return kwargs
+
+
 def rendition_allowed(filter_specs: str) -> bool:
     """Checks a given rendition filter is allowed"""
     allowed_filters = grapple_settings.ALLOWED_IMAGE_FILTERS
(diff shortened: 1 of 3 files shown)


grapple/types/images.py
```diff
@@ -181,6 +197,7 @@ def resolve_src_set(
         info: GraphQLResolveInfo,
         sizes: list[int],
         format: str | None = None,
+        preserve_svg: bool = False,
```
zerolab, line 200: Based on a conversation in the Wagtail Slack, I am very much inclined to make `preserve_svg` default to True here (and on the rendition)

ie. until Wagtail (well, Willlow) can actually rasterize SVGs, the only way to work with SVGs is to use `preserve_svg` in Grapple, and `preserve-svg` in regular templates, which adds so much extra effort.

What d'ya think?

zerolab (approved):
This all LGTM.

Left a provocation (well, a challenge question) and would love to hear your thought as an implementer

grapple/types/images.py
mgax: Yes, I think it very much makes sense. I can't imagine a scenario where you would _not_ want `preserveSvg: true` as a default, even if it were possible to rasterize them.

We need to make sure this doesn't break on Wagtail < 5 though; I remember the code was relying on the fact that, on old Wagtails, it would not be possible to change the default `False` value.

zerolab: I think it should be fine. as `preserve_svg` filters out non-SVG filters - https://github.com/torchbox/wagtail-grapple/blob/d8bb9222b70584aa65b9307bfbb7d6139170fbd8/grapple/types/images.py#L153-L161

mgax: Yes, but `to_svg_safe_spec` is imported only for Wagtail >= 5; just wanted to mention this, as there don't seem to be a CI matrix that includes older Wagtails: https://github.com/torchbox/wagtail-grapple/blob/d8bb9222b70584aa65b9307bfbb7d6139170fbd8/grapple/types/images.py#L24-L25

---
I can implement the change; do you feel it should be part of this PR or separate?

zerolab: we run tox with tox-gh-actions in CI, so the matrix is in https://github.com/torchbox/wagtail-grapple/blob/main/tox.ini

let's follow up in a separate PR. We'd want to wrap `if instance.is_svg() and preserve_svg: ` in an `if WAGTAIL_VERSION > (5, 0)` to be sure

mgax: > we run tox with tox-gh-actions in CI, so the matrix is in https://github.com/torchbox/wagtail-grapple/blob/main/tox.ini

@zerolab I think something is missing. I couldn't find a CI job that mentioned an old version of Wagtail so I made a branch to see if I can break it. AFAICT this should have failed CI: https://github.com/mgax/wagtail-grapple/commit/7b52a38c60fbcc4f2cec1bf9554cb82debf320b1

PR #380, opened 2023-12-12
Avoid using hardcoded localhost in Redirects

JakubMastalerz:
This PR changes `old_url` handling in `RedirectType` to replace hardcoded URL generation with dynamically generated URLs based on wagtail's `sites` properties.

Additionally:

- Fixes issue where query would fail if `Redirect` had no specified target.
- Introduces general test suite for redirects.

The changes were tested for performance and did not increase the number of queries when querying `Redirects`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,9 @@
 
 ### Changed
 
--   `new_url` field on `RedirectType` is no longer required ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
--   `old_url` field on `RedirectType` now returns url if site is specified and list of urls for all sites if not. ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
+-   Renamed `RedirectType` to `RedirectObjectType` for clarity ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
+-   `new_url` field on `RedirectObjectType` is no longer required ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
+-   `old_url` field on `RedirectObjectType` now returns url if site is specified and list of urls for all sites if not. ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
 
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,13 @@
 
 -   Factory for `Redirect` model in test app ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
 -   Test suite for `Redirect` behaviour ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
+-   `site` field on `RedirectObjectType` ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
 
 ### Changed
 
 -   Renamed `RedirectType` to `RedirectObjectType` for clarity ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
 -   `new_url` field on `RedirectObjectType` is no longer required ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
--   `old_url` field on `RedirectObjectType` now returns url if site is specified and list of urls for all sites if not. ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
+-   `RedirectsQuery` now returns a separate redirect object for each existing site when `site=None` ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
 
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,10 @@
 
 ### Added
 
--   Factory for `Redirect` model in test app ([tba](https://github.com/torchbox/wagtail-grapple/pull/tba)) @jakubmastalerz
+-   Factory for `Redirect` model in test app ([tba](https://github.com/torchbox/wagtail-grapple/pull/tba)) @JakubMastalerz
+-   Test suite for `Redirect` behaviour ([tba](https://github.com/torchbox/wagtail-grapple/pull/tba)) @JakubMastalerz
+
+
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,9 @@
 -   Factory for `Redirect` model in test app ([tba](https://github.com/torchbox/wagtail-grapple/pull/tba)) @JakubMastalerz
 -   Test suite for `Redirect` behaviour ([tba](https://github.com/torchbox/wagtail-grapple/pull/tba)) @JakubMastalerz
 
+### Changed
 
+-   `newUrl` field on `RedirectType` is no longer required ([tba](https://github.com/torchbox/wagtail-grapple/pull/tba)) @JakubMastalerz
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
 ## Unreleased
 
+### Added
+
+-   Factory for `Redirect` model in test app ([tba](https://github.com/torchbox/wagtail-grapple/pull/tba)) @jakubmastalerz
+
 ### Fixed
 
 - `test_src_set_invalid_format` not working with Wagtail 5.2 and above. ([#378](https://github.com/torchbox/wagtail-grapple/pull/378)) @JakubMastalerz
+
 
 ## [0.23.0] - 2023-09-29
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,17 +1,11 @@
 ## Unreleased
 
-### Added
-
--   Factory for `Redirect` model in test app ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
--   Test suite for `Redirect` behaviour ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
--   `site` field on `RedirectObjectType` ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
-
 ### Changed
 
--   Renamed `RedirectType` to `RedirectObjectType` for clarity ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
--   `new_url` field on `RedirectObjectType` is no longer required ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
--   `RedirectsQuery` now returns a separate redirect object for each existing site when `site=None` ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
-
+-   Renamed `RedirectType` to `RedirectObjectType` ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
+-   `RedirectsQuery` now returns a `Redirect` for each associated `Site` ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
+    -   Added `site` field on `RedirectObjectType`
+    -   Changed `new_url` field on `RedirectObjectType` to be optional
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,14 @@
 
 ### Added
 
--   Factory for `Redirect` model in test app ([tba](https://github.com/torchbox/wagtail-grapple/pull/tba)) @JakubMastalerz
--   Test suite for `Redirect` behaviour ([tba](https://github.com/torchbox/wagtail-grapple/pull/tba)) @JakubMastalerz
+-   Factory for `Redirect` model in test app ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
+-   Test suite for `Redirect` behaviour ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
 
 ### Changed
 
--   `newUrl` field on `RedirectType` is no longer required ([tba](https://github.com/torchbox/wagtail-grapple/pull/tba)) @JakubMastalerz
+-   `new_url` field on `RedirectType` is no longer required ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
+-   `old_url` field on `RedirectType` now returns url if site is specified and list of urls for all sites if not. ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
+
 
 ### Fixed
 
--- a/grapple/types/redirects.py
+++ b/grapple/types/redirects.py
@@ -10,7 +10,7 @@ from wagtail.models import Page, Site
 from .pages import get_page_interface
 
 
-class RedirectType(graphene.ObjectType):
+class RedirectObjectType(graphene.ObjectType):
     old_path = graphene.String(required=True)
     old_url = graphene.String(required=True)
     new_url = graphene.String(required=False)
@@ -50,9 +50,12 @@ class RedirectType(graphene.ObjectType):
         if self.redirect_page is not None:
             return self.redirect_page.specific
 
+    class Meta:
+        name = "Redirect"
+
 
 class RedirectsQuery:
-    redirects = graphene.List(graphene.NonNull(RedirectType), required=True)
+    redirects = graphene.List(graphene.NonNull(RedirectObjectType), required=True)
 
     # Return all redirects.
     def resolve_redirects(self, info, **kwargs):
--- a/grapple/types/redirects.py
+++ b/grapple/types/redirects.py
@@ -29,7 +29,7 @@ class RedirectObjectType(graphene.ObjectType):
         site and `old_path`.
         """
 
-        return f"{self.site.root_url}/{self.old_path}"
+        return self.site.root_url + self.old_path
 
     def resolve_new_url(self, info, **kwargs) -> Optional[str]:
         """
--- a/grapple/types/redirects.py
+++ b/grapple/types/redirects.py
@@ -7,6 +7,8 @@ import graphene
 from wagtail.contrib.redirects.models import Redirect
 from wagtail.models import Page, Site
 
+from grapple.types.sites import SiteObjectType
+
 from .pages import get_page_interface
 
 
@@ -15,6 +17,7 @@ class RedirectObjectType(graphene.ObjectType):
     old_url = graphene.String(required=True)
     new_url = graphene.String(required=False)
     page = graphene.Field(get_page_interface())
+    site = graphene.Field(SiteObjectType, required=False)
     is_permanent = graphene.Boolean(required=True)
 
     def resolve_old_url(self, info, **kwargs) -> str:
@@ -23,8 +26,9 @@ class RedirectObjectType(graphene.ObjectType):
         Otherwise, return a JSON string representing a list of all site urls.
         """
         if self.site:
-            return f"http://{self.site.hostname}:{self.site.port}/{self.old_path}"
+            return f"{self.site.root_url}/{self.old_path}"
 
+        # TODO: Remove after making site required
         if self.site is None:
             sites_QS = Site.objects.all()
             old_url_list = []
--- a/grapple/types/redirects.py
+++ b/grapple/types/redirects.py
@@ -1,7 +1,10 @@
+from typing import Optional
+
 import graphene
 
 from django.conf import settings
 from wagtail.contrib.redirects.models import Redirect
+from wagtail.models import Page
 
 from .pages import get_page_interface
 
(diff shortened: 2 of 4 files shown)


grapple/types/redirects.py
```diff
@@ -1,31 +1,52 @@
...
 import graphene
 
-from django.conf import settings
 from wagtail.contrib.redirects.models import Redirect
+from wagtail.models import Page, Site
 
 from .pages import get_page_interface
 
 
 class RedirectType(graphene.ObjectType):
```
kbayliss, line 13: Can you please rename this to `RedirectObjectType` to be consistent with other types? And perhaps make this nicer to use by specifying a name too:

```suggestion
class RedirectObjectType(graphene.ObjectType):
    ... fields ...
    
    class Meta:
        name = 'Redirect'
```

`RedirectType` isn't as clear.

grapple/types/redirects.py
```diff
@@ -1,31 +1,52 @@
...
-    # Give old_path with BASE_URL attached.
-    def resolve_old_url(self, info, **kwargs):
-        return settings.BASE_URL + self.old_path
+    def resolve_old_url(self, info, **kwargs) -> str:
+        """
+        If the redirect is for a specific site, append hostname and port to path.
+        Otherwise, return a JSON string representing a list of all site urls.
+        """
+        if self.site:
+            return f"http://{self.site.hostname}:{self.site.port}/{self.old_path}"
```
kbayliss, line 26: **Issue:** Not all sites will start with "http://", so this value is erroneous. Wagtail editors can specify "https://" for a `Site` by changing the port to 443.

grapple/types/redirects.py
```diff
@@ -1,31 +1,52 @@
...
 
-    # Give old_path with BASE_URL attached.
-    def resolve_old_url(self, info, **kwargs):
-        return settings.BASE_URL + self.old_path
+    def resolve_old_url(self, info, **kwargs) -> str:
+        """
+        If the redirect is for a specific site, append hostname and port to path.
+        Otherwise, return a JSON string representing a list of all site urls.
+        """
+        if self.site:
```
kbayliss, line 25: **Question:** What is `self.site` here? I don't see any field, attribute, or property matching `site`. What am I missing?

grapple/types/redirects.py
```diff
@@ -1,31 +1,52 @@
...
+            return f"http://{self.site.hostname}:{self.site.port}/{self.old_path}"
+
+        if self.site is None:
+            sites_QS = Site.objects.all()
```
kbayliss, line 29: **Issue:** Editors control what sites a redirect applies to from within Wagtail. So, every redirect created within Wagtail will either be assigned to a single site or "all sites" (which is the empty/none value). 

However, this code ignores that and just fetches all `Site`s in the database, negating editors' choice.

tests/testapp/factories.py
```diff
@@ -180,3 +182,20 @@ class Meta:
...
+
+class RedirectFactory(factory.django.DjangoModelFactory):
+    """
+    TODO: Consider moving to wagtail factories?
+    """
+
+    old_path = factory.Faker("slug")
+    # Note: Site needs `hostname` and `port` to be a unique combination,
+    # so this seems like the best option:
+    site = factory.Iterator(Site.objects.all())
```
kbayliss, line 195: While a nice idea, `wagtail-factories` contains a factory for Site's so we can just use that:

https://github.com/wagtail/wagtail-factories/blob/main/src/wagtail_factories/factories.py#L132

```suggestion
    site = factory.SubFactory(wagtail_factories.SiteFactory)```

File: CHANGELOG.md [hunk 6]
```diff
@@ -1,5 +1,16 @@
 ## Unreleased
 
+### Added
+
+-   Factory for `Redirect` model in test app ([#380](https://github.com/torchbox/wagtail-grapple/pull/380)) @JakubMastalerz
```
kbayliss, line 5: Did you try doing the below?

```suggestion
-   Factory for `Redirect` model in test app (!380) @JakubMastalerz
```

or

```suggestion
-   Factory for `Redirect` model in test app ([!380]) @JakubMastalerz
```

tests/test_redirects.py
```diff
@@ -0,0 +1,156 @@
+import json
+
+from test_grapple import BaseGrappleTest
+from testapp.factories import RedirectFactory
+from testapp.models import BlogPage
+from wagtail_factories import SiteFactory
+
+
+class TestRedirectQueries(BaseGrappleTest):
+    def setUp(self) -> None:
```
kbayliss, line 10: **Question:** Why `setUp` not `setUpTestData`? This looks like data that can be shared between all test methods within this class.

grapple/types/redirects.py
kbayliss, line 13: **Issue:** Are we missing a `site` field? `wagtail.contrib.redirects.models` contains a `site` field, so why doesn't our `ObjectType` contain one?

As an API consumer calling a GraphQL API using wagtail-grapple, how would I filter the redirects for a specific site? Note, this last point should be considered on top of adding a `site` field - perhaps a query would help?

grapple/types/redirects.py
```diff
@@ -1,31 +1,52 @@
...
     old_url = graphene.String(required=True)
-    new_url = graphene.String(required=True)
+    new_url = graphene.String(required=False)
     page = graphene.Field(get_page_interface())
     is_permanent = graphene.Boolean(required=True)
 
-    # Give old_path with BASE_URL attached.
-    def resolve_old_url(self, info, **kwargs):
-        return settings.BASE_URL + self.old_path
+    def resolve_old_url(self, info, **kwargs) -> str:
```
kbayliss, line 20: This method is doing a lot of fudging.

Can you reuse (i.e. call directly, not duplicate here) any code from: https://github.com/wagtail/wagtail/blob/main/wagtail/contrib/redirects/models.py - perhaps one of the properties/methods?

tests/testapp/factories.py
```diff
@@ -180,3 +182,20 @@ class Meta:
...
+
+
+class RedirectFactory(factory.django.DjangoModelFactory):
+    """
+    TODO: Consider moving to wagtail factories?
+    """
+
+    old_path = factory.Faker("slug")
+    # Note: Site needs `hostname` and `port` to be a unique combination,
+    # so this seems like the best option:
```
kbayliss, line 194: Let's rm this comment after changing the below to a `SiteFactory`

kbayliss (changes requested):
A great start, though there's some choices that will cause issues that need addressing.

grapple/types/redirects.py
JakubMastalerz: This site relates to a field in `Redirect` model. You can see similar use in the same file, like `self.link` and `self.redirect_page`, though I have to admit I'm a bit confused about why it works.

CHANGELOG.md
JakubMastalerz: Yes, didn't work unfortunately, so I reversed to the old method.

tests/test_redirects.py
JakubMastalerz: I believe it still is shared between all methods. I use the definitions from `setUp` in `test_new_url_sources`

tests/testapp/factories.py
JakubMastalerz: This comment is actually a response to issues arising from using `SiteFactory` there, but I will re-test it.

grapple/types/redirects.py
JakubMastalerz: Addressed in 6a7294b264f3584e04fdc0d279e96a9e8c048a37

grapple/types/redirects.py
JakubMastalerz: My understanding is that not specifying `site` means that a redirect is assigned to "all sites". In this case, we are getting all sites, to return a list of sites the redirect applies to.

grapple/types/redirects.py
```diff
@@ -1,37 +1,61 @@
...
 
-from django.conf import settings
 from wagtail.contrib.redirects.models import Redirect
+from wagtail.models import Page, Site
 
 from .pages import get_page_interface
 
 
-class RedirectType(graphene.ObjectType):
+class RedirectObjectType(graphene.ObjectType):
```
kbayliss, line 13: So that developers can more clearly see what `Site` this redirect applies to, can you add a `site` field to this?

grapple/types/redirects.py
```diff
@@ -1,37 +1,61 @@
...
     old_url = graphene.String(required=True)
-    new_url = graphene.String(required=True)
+    new_url = graphene.String(required=False)
     page = graphene.Field(get_page_interface())
     is_permanent = graphene.Boolean(required=True)
 
-    # Give old_path with BASE_URL attached.
-    def resolve_old_url(self, info, **kwargs):
-        return settings.BASE_URL + self.old_path
+    def resolve_old_url(self, info, **kwargs) -> str:
```
kbayliss, line 20: As discussed just now, let's simplify this.

Let's say you have a `Redirect` for two sites. Currently, you would return a single object with multiple values for `site` and `old_url`:

```
Redirect(
    old_url=["http://xx.com/hello", "http://yy.com/hello"]
    site=[x,y]
    ...
)
```

But this is confusing because the field names suggest singular values. Also, API consumers need to loop through `Redirect`s and then loop through each redirect's sites/URLs to set up redirect handling within their codebase, which is arduous.

To simplify this, can you return a separate object for each site/url? For example, the above response would become:

```
Redirect(
    old_url="http://xx.com/hello"
    site=x
    ...
)
Redirect(
    old_url="http://yy.com/hello"
    site=y
    ...
)
```

kbayliss (changes requested):

tests/test_redirects.py
kbayliss: `setUp` runs for each test within a test case, whereas `setUpTestData` is run once for the class and is available for all methods to use.

So while it seems like `setUp` is shared between methods within a class, it's actually redoing the work `n` times.

tests/testapp/factories.py
JakubMastalerz: Addressed in 2822f362c4cfc4dfd11385da06ac38d630f1824a

tests/testapp/factories.py
JakubMastalerz: Addressed in 2822f362c4cfc4dfd11385da06ac38d630f1824a

grapple/types/redirects.py
JakubMastalerz: Thanks! Addressed in 33561505891fe6ec7d6a5901f9301af1d26aaad5

tests/test_redirects.py
JakubMastalerz: Thanks, addressed in 33561505891fe6ec7d6a5901f9301af1d26aaad5

grapple/types/redirects.py
JakubMastalerz: Thanks, added a site field in 33561505891fe6ec7d6a5901f9301af1d26aaad5

grapple/types/redirects.py
JakubMastalerz: Addressed in 33561505891fe6ec7d6a5901f9301af1d26aaad5

grapple/types/redirects.py
JakubMastalerz: Implemented this in 75707b2fcdd68d10feccadbeca9a62755fe8bade

When returning a list of redirects, for redirects with `site=None` we now create a copy for each existing site and then remove the one with no site from the list.

grapple/types/redirects.py
```diff
@@ -1,38 +1,76 @@
...
+    class Meta:
+        name = "Redirect"
+
 
 class RedirectsQuery:
-    redirects = graphene.List(graphene.NonNull(RedirectType), required=True)
+    redirects = graphene.List(graphene.NonNull(RedirectObjectType), required=True)
 
     # Return all redirects.
     def resolve_redirects(self, info, **kwargs):
```
kbayliss, line 58:
```suggestion
    def resolve_redirects(self, info, **kwargs) -> list[Redirect]:
```

tests/test_redirects.py
```diff
@@ -0,0 +1,211 @@
...
+        super().setUpTestData()
+        cls.page = BlogPage(title="Test page", slug="test-page-url", date="2020-01-01")
+        cls.home.add_child(instance=cls.page)
+
+    def test_basic_query(self):
+        """
+        Test that Redirect fields can be queried through graphql.
+        """
+
+        self.redirect = RedirectFactory(
```
kbayliss, line 19: By using `self`, you are modifying data held in the class, which is not what you want here. You are creating something for use just within this method, which is achieved by not including `self`, e.g.:

```
redirect = RedirectFactory(...)
```

However, you are not reusing `redirect` anywhere, so there's no need to assign it to a variable. Instead, you can just do:

```
RedirectFactory(...)
```

tests/test_redirects.py
```diff
@@ -0,0 +1,211 @@
...
+                oldUrl
+            }
+        }
+        """
+
+        result = self.client.execute(query)["data"]["redirects"]
+
+        self.assertEqual(result[0]["oldUrl"], "http://another-test-site:82/old-path")
+        self.assertEqual(result[1]["oldUrl"], "http://localhost/old-path")
+        self.assertEqual(result[2]["oldUrl"], "http://test-site:81/old-path")
```
kbayliss, line 211: There are some additional checks that would be beneficial to add to this test to ensure the data is in the correct format before comparing values.

Here are the comments I would find helpful when reading the code:

```suggestion
    def test_all_sites_url(self):
        """
        Test that when no site is specified on a redirect, that redirect is
        shown for each existing site.
        """

        # Create two additional sites (on top of the default `Localhost` one).
        SiteFactory(hostname="test-site", port=81)
        SiteFactory(hostname="another-test-site", port=82)

        # Create a `Redirect` without a `site`.
        RedirectFactory(
            old_path="old-path",
            site=None,
        )

        query = """
        {
            redirects {
                oldUrl
            }
        }
        """

        result = self.client.execute(query)["data"]["redirects"]

        # Sanity check to ensure there are three `Site`s.
        self.assertEqual(Site.objects.count(), 3)
        
        # Sanity check to ensure only one `Redirect` is in the database.
        self.assertEqual(Redirect.objects.count(), 1)

        # Ensure there are three `Redirect`s returned despite only 1 object
        # in the database.
        self.assertEqual(len(result), 3)

        self.assertEqual(result[0]["oldUrl"], "http://another-test-site:82/old-path")
        self.assertEqual(result[1]["oldUrl"], "http://localhost/old-path")
        self.assertEqual(result[2]["oldUrl"], "http://test-site:81/old-path")
```

Can you experiment with this and see if any other tests would benefit from additional checks (like checking the number of results)?

grapple/types/redirects.py
```diff
@@ -1,38 +1,76 @@
...
+        for redirect in redirect_list:
+            if redirect.site is None:
+                for site in sites_QS:
+                    new_redirect = copy.copy(redirect)
+                    new_redirect.site = site
+                    redirect_list.append(new_redirect)
+
+        redirect_list = filter(lambda x: (x.site is not None), redirect_list)
+
+        return redirect_list
```
kbayliss, line 76: Can slightly reduce the complexity here by doing:

```suggestion
        redirects = Redirect.objects.select_related("redirect_page").all()
        finalised_redirects: list[Redirect] = []  # Redirects to return within API response.

        for redirect in redirects:
            if redirect.site is None:
                # Duplicate Redirect for each Site as it applies to all Sites.
                for site in Site.objects.all():
                    _new_redirect = copy.copy(redirect)
                    _new_redirect.site = site
                    finalised_redirects.append(_new_redirect)
            else:
                finalised_redirects.append(redirect)

        return finalised_redirects
```

kbayliss (changes requested):

grapple/types/redirects.py
```diff
@@ -1,38 +1,76 @@
...
 
 
-class RedirectType(graphene.ObjectType):
+class RedirectObjectType(graphene.ObjectType):
     old_path = graphene.String(required=True)
     old_url = graphene.String(required=True)
-    new_url = graphene.String(required=True)
+    new_url = graphene.String(required=False)
     page = graphene.Field(get_page_interface())
+    site = graphene.Field(SiteObjectType, required=False)
```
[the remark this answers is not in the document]
JakubMastalerz: It could be, though I'm not sure if that would be clear for users given that `site` is not required for `redirects`. Maybe throwing a custom exception instead would be better?

grapple/types/redirects.py
```diff
@@ -1,38 +1,76 @@
...
     is_permanent = graphene.Boolean(required=True)
 
-    # Give old_path with BASE_URL attached.
-    def resolve_old_url(self, info, **kwargs):
-        return settings.BASE_URL + self.old_path
+    def resolve_old_url(self, info, **kwargs) -> str:
```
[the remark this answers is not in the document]
JakubMastalerz: The return type should be `-> str` because `site` is now required, and as such this function will not return `None`. Addressed in cba26966a00bbfb670c390908a84cd91b8cfb4c2

grapple/types/redirects.py
```diff
@@ -1,38 +1,76 @@
...
+        """
+        Resolve the value of `old_url` using the `root_url` of the associated
+        site and `old_path`.
+        Note: `self.site` should never be none because of `resolve_redirects`.
+        """
+        if self.site:
+            return f"{self.site.root_url}/{self.old_path}"
+
+        if self.site is None:
+            return None
```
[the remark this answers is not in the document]
JakubMastalerz: You're right, this shouldn't be possible, changed in cba26966a00bbfb670c390908a84cd91b8cfb4c2

grapple/types/redirects.py
JakubMastalerz: Addressed in cba26966a00bbfb670c390908a84cd91b8cfb4c2

grapple/types/redirects.py
JakubMastalerz: Thanks! Implemented that in cba26966a00bbfb670c390908a84cd91b8cfb4c2

tests/test_redirects.py
JakubMastalerz: Thanks! Improved documentation of `test_redirects.py` in 40eeb98e50bd5cf9895d758fbd761f3f2c474567

tests/test_redirects.py
JakubMastalerz: Addressed in 40eeb98e50bd5cf9895d758fbd761f3f2c474567

grapple/types/redirects.py
```diff
@@ -1,38 +1,72 @@
...
+        """
+        Resolve the value of `old_url` using the `root_url` of the associated
+        site and `old_path`.
+        """
 
-    # Get new url
-    def resolve_new_url(self, info, **kwargs):
-        if self.redirect_page is None:
-            return self.link
+        return f"{self.site.root_url}/{self.old_path}"
```
kbayliss, line 29: **Issue:** This always results in a double slash after the hostname because `old_path` enforces that the path begins with `/`. 

This is tested using Wagtail v5.1.2, so may not be true for other versions.

Can you check? We may be able to change this to:

```suggestion
        return self.site.root_url + self.old_path
```

tests/test_redirects.py
```diff
@@ -0,0 +1,246 @@
...
+        # Create a redirect with both `redirect_link` and `redirect_page`.
+        RedirectFactory(redirect_link="/test/incorrect", redirect_page=self.page)
+        # Create a redirect with just `redirect_page`.
+        RedirectFactory(
+            redirect_link="",
+            is_permanent=True,
+            redirect_page=self.page,
+        )
+        # Create a redirect with just `redirect_link`.
+        RedirectFactory(redirect_link="http://test/just-link", redirect_page=None)
```
kbayliss, line 112: **Issue:** This is not a valid URL - can we make it so?

grapple/types/redirects.py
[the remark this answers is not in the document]
kbayliss: We're not saying it's required for `Redirect` instances though, we're saying it's required when querying for them via GraphQL (which it is as `RedirectsQuery` makes sure of that)

grapple/types/redirects.py
kbayliss: Does this work locally for you when running `python manage.py test`? If so, what python version are you using?

Using Python v3.8.11, I get a `TypeError` (_TypeError: 'type' object is not subscriptable_) unless I change to:

```suggestion  
    def resolve_redirects(self, info, **kwargs) -> List[Redirect]:
```

(and importing `List` from `typing`)

kbayliss:
@JakubMastalerz this is looking GREAT! As discussed earlier, can you review the performance of these changes to check we're not adversely affecting performance? In particular, I'm interested in the number of queries being made to ensure we're not causing any N+1 queries.

grapple/types/redirects.py
JakubMastalerz: I'm using Python v3.11.2. It works locally for me, though changing it to `-> List[Redirects]` works as well.

grapple/types/redirects.py
JakubMastalerz: I was using Wagtail v5.2.1. I see the `normalize_path` function in the Redirect model, and it should be adding that `/`, but when I run tests the `old_url` is generated in the correct way.

grapple/types/redirects.py
```diff
@@ -1,38 +1,72 @@
...
+        return self.link
 
     # Return the page that's being redirected to, if at all.
-    def resolve_page(self, info, **kwargs):
+    def resolve_page(self, info, **kwargs) -> Optional[Page]:
         if self.redirect_page is not None:
             return self.redirect_page.specific
 
+    class Meta:
+        name = "Redirect"
```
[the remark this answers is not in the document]
JakubMastalerz: Addressed in c2a53657cc16941c4ea4a6307f0e21fc27809e99

grapple/types/redirects.py
JakubMastalerz: Went with your suggestion, since it seems to work with more Python versions. c2a53657cc16941c4ea4a6307f0e21fc27809e99

grapple/types/redirects.py
JakubMastalerz: Addressed this in 6950794a844166aa5bb9300f3a51ea3ea6b9143b.

kbayliss (approved):

PR #385, opened 2024-03-27
Add py312, dj5, wag6 to tox matrix

jams2:

PR #387, opened 2024-04-10
Support Wagtail 5.2+, Wagtail 6+, Python 3.12 and Django 5

Morsey187:
This PR adds support for: 
  - Wagtail 6.0
  - Wagtail 5.2
  - Djangov5
  - Pyhon 3.12

There was one fix needed for Wagtail v6.0 to resolve an import error [Query model moved to wagtail.contrib.search_promotions](https://docs.wagtail.org/en/stable/releases/6.0.html#query-model-moved-to-wagtail-contrib-search-promotions), and I've included commits from https://github.com/torchbox/wagtail-grapple/pull/385 for added Python and Django coverage.

### Notes 

You can test individual environments with tox using:
```bash
tox -e py310-django42-wagtail60
```


grapple/utils.py
```diff
@@ -4,14 +4,26 @@
...
 from .types.structures import BasePaginatedType, PaginationType
 
 
+if grapple_settings.ADD_SEARCH_HIT:
+    if WAGTAIL_VERSION >= (6, 0):
+        try:
+            from wagtail.contrib.search_promotions.models import Query
+        except ImportError as e:
+            raise ImportError(
+                "wagtail.contrib.search_promotions app is required for Wagtail 6.0+"
```
Morsey187, line 21: **Open to suggestions here** 

The Query model is only used if `grapple_settings.ADD_SEARCH_HIT` is set to True and an explict error here seemed appropriate to help users when upgrading.

tests/test_blog.py
```diff
@@ -500,7 +500,8 @@ def test_blog_embed(self):
             "height": 113,
             "html": '<iframe width="200" height="113" src="https://www.youtube.com/embed/_U79Wc965vw?feature=oembed" '
             'frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; '
-            'picture-in-picture; web-share" allowfullscreen title="Wagtail Space 2018"></iframe>',
+            'picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen '
```
Morsey187, line 503: Needed to fix tests, youtube must have made an update.

tests/testapp/migrations/0001_initial.py
```diff
@@ -327,7 +327,6 @@ class Migration(migrations.Migration):
                     models.ImageField(
```
Morsey187, line 327: Issue:
My local build wants to remove this field, although, I'm not sure why as the field looks to exist still
https://github.com/wagtail/wagtail/blob/main/wagtail/images/models.py#L246


Would be great if this could be tested by someone else locally, to see if migration changes are needed and/or requested.

tox.ini
```diff
@@ -45,9 +48,12 @@ deps =
     django32: Django>=3.2,<3.3
     django41: Django>=4.1,<4.2
```
zerolab, line 49: would be good to drop Django < 4.2 and Wagtail < 5.2

grapple/utils.py
jams2: Maybe `ImproperlyConfigured` from Django's exceptions would be a better fit than `ImportError`. I would also suggest making the message something like "wagtail.contrib.search_promotions must be installed if grapple_settings.ADD_SEARCH_HIT=True and using Wagtail >= 6.0"

tests/testapp/migrations/0001_initial.py
jams2: Are you referring to the change below (line 330)? It's just removing the verbose name attribute, which makes sense as it matches the field name. If I run `makemigrations` against your branch I get the following output:

```
Migrations for 'testapp':
  testapp/migrations/0003_alter_blogpage_body_alter_customimage_file.py
    - Alter field body on blogpage
    - Alter field file on customimage
```

So it seems there's some changes that haven't been added to migrations. In the past, for projects with test apps in the past I've removed all the test app migrations and remade them fresh using the newest supported versions of Django and Wagtail - you just need to run tox on all the envs to make sure the migrations don't fail on any particular version.

tox.ini
Morsey187: That sound sensible, for clarity are you suggesting the following:
```ini
# File: tox.ini
##########

envlist =
    py{38,39,310,311,312}-django{42,50}-wagtail{52,60}
    interactive


...
    django42: Django>=4.2,<5.0
    django50: Django>=5.0,<5.1
    wagtail52: wagtail>=5.2,<6.0
    wagtail60: wagtail>=6.0,<6.1
    interactive: wagtail>=5.2
```
Note: I referenced https://docs.wagtail.org/en/stable/releases/upgrading.html#compatible-django-python-versions to decide on the envlist version mapping.

Curious to understand how support is decided for Wagtail though, it looks like Django has dropped long term support for 4.1 in December, however, how do we decide on what versions of Wagtail to support?
Wouldn't it make more sense to support Wagtail < 5.0 since it's compatible with Django 4.2?

Would we also want to suppor Python 3.7?

zerolab: What versions one supports is a mixed bag of
- guidance from https://github.com/wagtail-nest/
- officially supported versions
  - https://endoflife.date/python (we can see clearly that 3.7 is out of the picture, and frankly we should think of starting to drop 3.8 too)
  - https://endoflife.date/django
  - https://endoflife.date/wagtail

in an ideal world we'd support the widest range, but then that adds maintenance burder, increases the CI time and resource waste, and doesn't help push implementers to keep up to date.

I talked about my approach at our internal tech conference in October

tox.ini
```diff
@@ -2,10 +2,7 @@
 min_version = 4.0
 
 envlist =
-    py{38,39,310}-django{32,41}-wagtail{41,50,51}
-    py{311}-django{41,42}-wagtail{50,51}
-    py{311,312}-django{41,42}-wagtail{50,51,52}
-    py{311,312}-django{42,50}-wagtail60
+    py{38,39,310,311,312}-django{42,50}-wagtail{52,60}
```
zerolab, line 5: Can you add Python 3.12, Wagtail 6, and remove Wagtail 4 from PyPI classifiers https://github.com/torchbox/wagtail-grapple/blob/main/pyproject.toml#L17 ?

tox.ini
Morsey187: Nice, thanks for the response Dan, I'll also see about watching your talk, I was ill that day but have a link to the recordings.

I've made a slight change to the orginal envlist since my inital comment, as it looks like Django doesn't support python 3.8 and 3.9 https://docs.djangoproject.com/en/5.0/faq/install/#what-python-version-can-i-use-with-django

Otherwise I think this should be fine now, sorry I've made a mess of the commits trying to fix CI errors, I can spend time cleaning them up if that will help with reviewing.

tests/testapp/migrations/0001_initial.py
Morsey187: As discussed over slack, I've updated the migrations on django 4.2 and wagtail 5.2

grapple/utils.py
Morsey187: Completely agree! thanks both.

tests/requirements.txt
```diff
@@ -3,7 +3,7 @@ black>=23.9.1
 ruff==0.0.287
 
 # Runtime requirements
-Django>=3.2,<5.0
+Django>=3.2,<5.1
```
jams2, line 6: Should this be 4.2?

zerolab, line 6:
```suggestion
Django>=4.2,<5.1
```

pyproject.toml
```diff
@@ -26,13 +26,13 @@ classifiers = [
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
```
zerolab, line 27:
```suggestion
    "Programming Language :: Python :: 3.11",
    "Programming Language :: Python :: 3.12",
```

PR #393, opened 2024-05-10
Ensure ruff version consistency

### LINKED ISSUE #392: Inconsistant Ruff versions between CI pipeline and test app requirements.
It seems like Ruff versions used for linting in local development and in the pipeline are inconsistent. This leads to issues with formatting that aren't picked up by `pre-commit`, which runs on the local ruff, but cause the pipeline to fail.

The ruff version in `tests/requirements.txt` is `ruff==0.0.287`.

Ruff in the `.github/workflows/ruff.yml` workflow is installed as `python -Im pip install --user ruff`, which I believe installs the newest compatible version of ruff.

This leads to ruff check failing on PRs like https://github.com/torchbox/wagtail-grapple/pull/374

I think it would be beneficial to upgrade to the newest release of ruff (v0.4.4) and make sure that this is consistent between the requirements and the pipeline.

zerolab:
Fixes #392

### DIFF
--- a/.github/workflows/ruff.yml
+++ b/.github/workflows/ruff.yml
@@ -15,7 +15,8 @@ jobs:
     steps:
     - uses: actions/checkout@v4
 
-    - run: python -Im pip install --user ruff
+    # keep in sync with .pre-commit-config.yaml
+    - run: python -Im pip install --user ruff==0.4.4
 
     - name: Run ruff
       run: ruff --output-format=github grapple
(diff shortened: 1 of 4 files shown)


jams2:
Question: why not add a ruff requirement in pyproject.toml?

zerolab:
Because:
- we lint in CI with pre-commit.ci
- locally we encourage pre-commit
- the ruff.yml GitHub Action is for convenience and I don't want to install extra deps

PR #399, opened 2024-07-29
Fix issue with querying value on EmbedBlock

### LINKED ISSUE #395: resolve_value missing from EmbedBlock
When using the value field of an EmbedBlock in a query, it errors with `"Cannot return null for non-nullable field EmbedBlock.value."`

**Steps to reproduce:**

1. Define a StructBlock with an EmbedBlock field and register it with @register_streamfield_block

```
from wagtail.embeds import blocks as embeds_blocks

@register_streamfield_block
class EmbedStructBlock(blocks.StructBlock):
    embed = embeds_blocks.EmbedBlock()
```

2. Register the field for GraphQL

```
from grapple.types import streamfield as grapple_streamfield

    graphql_fields = [
        grapple_streamfield.EmbedBlock(),
    ]
```

3. Use this EmbedStructBlock in a StreamField of a Page model and execute a GraphQL query which requests ‘value’ on the embed field

```
… on EmbedStructBlock {
	testEmbed {
		value
	}
}
```

Switching the `resolve_raw_value` method of the `EmbedBlock` to `resolve_value` has solved this issue for me in the project I am currently working on

```
    def resolve_value(self, info, **kwargs):
        if isinstance(self, EmbedValue):
            return self
        return StreamFieldInterface.resolve_raw_value(info, **kwargs)
```

Should `resolve_raw_value` actually be `resolve_value`?

JakubMastalerz:
Fixes #395 

This PR changes the incorrectly named `resolve_raw_value` function to `resolve_value` on `EmbedBlock`. This fixes an issue where it was impossible to query for `value` on embeds using wagtail_grapple.

### DIFF
--- a/grapple/types/streamfield.py
+++ b/grapple/types/streamfield.py
@@ -363,6 +363,11 @@ class EmbedBlock(graphene.ObjectType):
             return self
         return StreamFieldInterface.resolve_raw_value(info, **kwargs)
 
+    def resolve_raw_value(self, info, **kwargs):
+        if isinstance(self, EmbedValue):
+            return self
+        return StreamFieldInterface.resolve_raw_value(info, **kwargs)
+
     def resolve_embed(self, info, **kwargs):
         embed = get_embed_object(self)
         if embed:
--- a/grapple/types/streamfield.py
+++ b/grapple/types/streamfield.py
@@ -1,5 +1,7 @@
 import inspect
 
+from typing import Optional
+
 import graphene
 
 # TODO: use specific imports
@@ -355,25 +357,23 @@ class EmbedBlock(graphene.ObjectType):
     class Meta:
         interfaces = (StreamFieldInterface,)
 
-    def resolve_url(self, info, **kwargs):
+    def resolve_url(self: EmbedValue, info, **kwargs) -> str:
         return get_embed_url(self)
 
-    def resolve_value(self, info, **kwargs):
+    def resolve_raw_value(self: EmbedValue, info, **kwargs) -> str:
         if isinstance(self, EmbedValue):
             return self
         return StreamFieldInterface.resolve_raw_value(info, **kwargs)
 
-    def resolve_raw_value(self, info, **kwargs):
-        if isinstance(self, EmbedValue):
-            return self
-        return StreamFieldInterface.resolve_raw_value(info, **kwargs)
+    def resolve_value(self: EmbedValue, info, **kwargs) -> str:
+        return EmbedBlock.resolve_raw_value(self, info, **kwargs)
 
-    def resolve_embed(self, info, **kwargs):
+    def resolve_embed(self: EmbedValue, info, **kwargs) -> Optional[str]:
         embed = get_embed_object(self)
         if embed:
             return embed.html
 
-    def resolve_raw_embed(self, info, **kwargs):
+    def resolve_raw_embed(self: EmbedValue, info, **kwargs) -> Optional[str]:
         embed = get_embed_object(self)
         if embed:
             return {
--- a/grapple/types/streamfield.py
+++ b/grapple/types/streamfield.py
@@ -358,7 +358,7 @@ class EmbedBlock(graphene.ObjectType):
     def resolve_url(self, info, **kwargs):
         return get_embed_url(self)
 
-    def resolve_raw_value(self, info, **kwargs):
+    def resolve_value(self, info, **kwargs):
         if isinstance(self, EmbedValue):
             return self
         return StreamFieldInterface.resolve_raw_value(info, **kwargs)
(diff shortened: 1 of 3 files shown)


tests/test_blog.py
```diff
@@ -506,11 +506,11 @@ def test_blog_embed(self):
         for block in body:
             if block["blockType"] == "VideoBlock":
                 embed = block["youtubeLink"]
-                expected_raw_value = "<div>\n    {}\n</div>\n".format(embed["embed"])
+                expected_value = "<div>\n    {}\n</div>\n".format(embed["embed"])
                 self.assertTrue(isinstance(embed["url"], str))
                 self.assertEqual(embed["embed"], raw_embed["html"])
                 self.assertEqual(embed["rawEmbed"], json.dumps(raw_embed))
-                self.assertEqual(embed["rawValue"], expected_raw_value)
```
zerolab, line 513: hmm.. I wonder how this hasn't failed before 🤔

JakubMastalerz: I'm not sure of the inner workings of GraphQL here. Is it possible it was working because `resolve_raw_value` existed, even though `raw_value` wasn't a specified field?

JakubMastalerz: Actually, from inspecting `EmbedBlock` in graphiQL, it seems that raw_value exists on it. Seems like it's inherited from `StreamFieldInterface`.

zerolab: so the task here then is to update the block to have a the proper `resolve_value` or have the tests for it

JakubMastalerz: I'm not sure what we would consider to be a good "value" for the `value` field. Since `raw_value`, returning HTML is available on `EmbedBlock`.

From inspecting the `wagtailembeds.embed`, it seems that the fields stored in the database are: id, URL, max_width, type, HTML, title, author_name, thumbnail_url, provider_name, width and height. 

I feel like these are well covered by the queryable fields we provide in wagtail-grapple, so I'm not what we'd like the `value` field to be.

What do you think?

cc @jams2

zerolab: in an ideal world, `value` would return the HTML snippet, whereas `raw_value` would return embed dict as JSON. This would be a breaking change, so for now let's get `value` and `raw_value` to return the HTML.

JakubMastalerz: In that case, are you happy with the implementation in c7e419c7f6cc73169066eb509743363719734a79?

`resolve_value` will return `EmbedValue` which is displayed as HTML or fall back to `raw_value`.
```
# grapple.types.streamfield

def resolve_value(self, info, **kwargs):
        if isinstance(self, EmbedValue):
            return self
        return StreamFieldInterface.resolve_raw_value(info, **kwargs)


# wagtailembeds.blocks

class EmbedValue:
   
    ...

    @cached_property
    def html(self):
        return embed_to_frontend_html(self.url, self.max_width, self.max_height)
def __str__(self):
        return self.html
```

`raw_value` and `resolve_raw_value` are still available on `EmbedBlock` as it's inherited from the `StreamFieldInterFace`


If this approach is okay then I would update the test to check for both `raw_value` and `value`.

zerolab: yeah, go for it

JakubMastalerz: Hi @zerolab, I have added a test for `raw_value` in 4f1682afa05425511d99d3d026d858ea3ac00a74. During testing, an issue came up, and it seems like the inherited `StreamFieldInterface.resolve_raw_value()` doesn't work on its own for the `EmbedBlock`.

I have added `resolve_raw_value()` back in to address this. It creates code repetition, which isn't ideal, but it seems to be the best working solution for now.

For added context, this is the error returned when attempting to query the `EmbedBlock` without overwriting `resolve_raw_value()`:
```
'errors': [{'message': "'EmbedValue' object has no attribute 'value'", 'locations': [{'line': 12, 'column': 33}], 'path': ['page', 'body', 11, 'youtubeLink', 'rawValue']}]
```

Let me know if you're okay with keeping both `resove_value()` and `resolve_raw_value()` on `EmbedBlock` for now.

grapple/types/streamfield.py
```diff
@@ -358,6 +358,11 @@ class Meta:
     def resolve_url(self, info, **kwargs):
         return get_embed_url(self)
 
+    def resolve_value(self, info, **kwargs):
+        if isinstance(self, EmbedValue):
+            return self
+        return StreamFieldInterface.resolve_raw_value(info, **kwargs)
+
```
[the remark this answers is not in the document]
zerolab: I keep forgetting this, but you got the gist of it. Type hints are most welcome, thanks

zerolab (approved):

PR #401, opened 2024-08-30
fix: rendition urls are not FQDNs

dopry:
Rendition URLs are returning only a path starting in 0.20.0 This updates it to return the same fully qualified domain name as the Image URLs.  I would appreciate it if this could be backported to at least 0.20.0 to create 0.20.1 since I'm still on the 0.20.0 branch. I'm running off a fork in the mean time so not operationally impacted.

### DIFF
--- a/grapple/types/images.py
+++ b/grapple/types/images.py
@@ -99,7 +99,7 @@ class ImageRenditionObjectType(DjangoObjectType):
     def resolve_url(
         instance: WagtailImageRendition, info: GraphQLResolveInfo, **kwargs
     ):
-        return instance.full_url
+        return get_media_item_url(instance)
 
 
 class ImageObjectType(DjangoObjectType):


dopry:
@zerolab I figured out what happened here. I didn't have WAGTAILADMIN_BASE_URL set, so my full_url wasn't generating a full url. I was still setting just BASE_URL.  Core has dropped the use of BASE_URL, so maybe we should in grapple a this point.  assuming full_url is working for both images and other things.  Although I still think it should be using MEDIA_URL. I haven't looked into the construction of full_url. I'll look into it on tuesday.

dopry:
Actually I take that back. I had a custom Rendition object that specified 

```
    graphql_fields = (
        GraphQLInt("id", required=True),
        GraphQLString("url", required=True),
        GraphQLString("width", required=True),
        GraphQLString("height", required=True),
        GraphQLImage("image", required=True),
        GraphQLString("file", required=True),
    )
```

and seemed to be masking the default object type.

PR #404, opened 2024-09-16
Fix circular import for `PageInterface` when using custom page interface

mgax:
I'd like to define a custom `PageInterface` (using the [`PAGE_INTERFACE `](https://wagtail-grapple.readthedocs.io/en/latest/getting-started/settings.html#page-interface) setting). Because `grapple.types.pages.PageInterface` is quite complex, I'd like to inherit from it, and add some additional fields.

This results in a circular import, because the definition of the `Page` class in the same module calls `get_page_interface`: https://github.com/torchbox/wagtail-grapple/blob/d9cd8f272836fbf914485d6af4b6c19c9b1589b1/grapple/types/pages.py#L165-L173

---
This PR moves `PageInterface` and `get_page_interface` to a new module, `interfaces.py`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,6 +106,7 @@ jobs:
           path: tests/.coverage*
           if-no-files-found: ignore
           retention-days: 1
+          include-hidden-files: true
 
   coverage:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,7 @@ jobs:
           path: tests/.coverage*
           if-no-files-found: ignore
           retention-days: 1
+          include-hidden-files: true
 
   tests-postgres:
     runs-on: ubuntu-latest
--- a/grapple/utils.py
+++ b/grapple/utils.py
@@ -248,3 +248,38 @@ def get_media_item_url(cls):
     if url[0] == "/":
         return settings.BASE_URL + url
     return url
+
+
+def serialize_struct_obj(obj):
+    rtn_obj = {}
+
+    if hasattr(obj, "raw_data"):
+        rtn_obj = []
+        for field in obj[0]:
+            rtn_obj.append(serialize_struct_obj(field.value))
+    # This conditionnal and below support both Wagtail >= 2.13 and <2.12 versions.
+    # The "stream_data" check can be dropped once 2.11 is not supported anymore.
+    # Cf: https://docs.wagtail.io/en/stable/releases/2.12.html#stream-data-on-streamfield-values-is-deprecated
+    elif hasattr(obj, "stream_data"):
+        rtn_obj = []
+        for field in obj.stream_data:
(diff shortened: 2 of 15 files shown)


mgax:
CI is failing for the _coverage_ job, but it looks like all recent CI runs are failing on that.

zerolab:
the `actions/upload-artifact` step needs `include-hidden-files: true` ([example](https://github.com/wagtail/Willow/blob/main/.github/workflows/test.yml#L59), [reference](https://github.blog/changelog/2024-08-19-notice-of-upcoming-deprecations-and-breaking-changes-in-github-actions-runners/))

tests/test_interfaces.py
```diff
@@ -7,7 +7,7 @@
 from testapp.factories import BlogPageFactory, CustomInterfaceBlockFactory
 from testapp.interfaces import CustomInterface
 
-from grapple.types.pages import PageInterface, get_page_interface
+from grapple.types.interfaces import PageInterface, get_page_interface
```
zerolab, line 10: can you update `testapp.interfaces.CustomInterface` to inherit from `PageInterface`?

we have `test_get_page_interface_with_custom_page_interface` but it was not fully testing your scenario

zerolab:
Since we're making these changes, it may be worth moving the `StreamFieldInterface` in the new interfaces module

.github/workflows/ci.yml
```diff
@@ -56,6 +56,7 @@ jobs:
           path: tests/.coverage*
           if-no-files-found: ignore
           retention-days: 1
+          include-hidden-files: true
```
zerolab, line 59: there's another one below (line 100something)

tests/test_interfaces.py
mgax: I've created a new interface, `CustomPageInterface`, for this purpose; `CustomInterface` is applied to types that are not pages and so things break.

.github/workflows/ci.yml
mgax: Ah, missed that, thanks!

mgax:
> Since we're making these changes, it may be worth moving the `StreamFieldInterface` in the new interfaces module

Ah, I meant to do that, but got carried away. Thanks!

grapple/utils.py
```diff
@@ -248,3 +248,38 @@ def get_media_item_url(cls):
...
+    rtn_obj = {}
+
+    if hasattr(obj, "raw_data"):
+        rtn_obj = []
+        for field in obj[0]:
+            rtn_obj.append(serialize_struct_obj(field.value))
+    # This conditionnal and below support both Wagtail >= 2.13 and <2.12 versions.
+    # The "stream_data" check can be dropped once 2.11 is not supported anymore.
+    # Cf: https://docs.wagtail.io/en/stable/releases/2.12.html#stream-data-on-streamfield-values-is-deprecated
+    elif hasattr(obj, "stream_data"):
```
zerolab, line 263: note to self: follow up with clean-up PR

zerolab (approved):
LGTM

Issue #386, opened 2024-04-10
Snippets don't identify themselves


jams2:
`SnippetObjectType` is implemented as a union of a project's registered snippet models. Unless the snippet classes themselves implement a field that can discriminate the specific type of a snippet instance, it can be difficult to determine which type of snippet you have in a client.

For example, if using a stream field with `SnippetChooserBlock`, I might query it as follows:

```graphql
query MyPages {
  pages(id: 28) {
    title
    ... on MyPage {
      streamFieldContent {
        ... on SnippetChooserBlock {
          id
          blockType
          snippet {
            ... on FooSnippet {
              fooField
            }
            ... on BarSnippet {
              barField
            }
          }
        }
      }
    }
  }
}
```

The issue then is determining exactly which snippet type I get in response. I could query `__typename` on `SnippetObjectType`, but this won't work in environments where introspection queries are forbidden.

To get around this in my project we've added our own snippet interface type, which has a `snippetType` field, which returns the snippet class name. Something similar would be a valuable addition to `grapple`.

PR #405, opened 2024-09-20
Replace SnippetObjectType with SnippetInterface

mgax:
Fixes https://github.com/torchbox/wagtail-grapple/issues/386
    
API clients could use a field on snippet objects to determine the type of snippet they are looking at. Therefore, we change the snippet type to an interface, similar to the page interface, so it can expose a new field called `snippetType`.

This PR includes a couple of refactoring commits which might be easier to view separately:
1. **Rename CustomInterface to AdditionalInterface**: The name `CustomInterface` is too similar with `CustomPageInterface` and the upcoming `CustomSnippetInterface` and their custom interfaces settings file. Better rename it to `AdditionalInterface` so there is less confusion.
2. **Rename the custom interface settings file**: The settings file will contain an override for the default snippet interface, in addition to the default page interface, so we're renaming it up front here.


grapple/types/interfaces.py
```diff
@@ -209,3 +209,18 @@ def resolve_raw_value(self, info, **kwargs):
             return self.value.source
 
         return self.value
+
+
+def get_snippet_interface():
+    return import_string(grapple_settings.SNIPPET_INTERFACE)
```
zerolab, line 215: I keep circling to the idea that we want the custom interfaces to inherit from our base ones (be it page, snippet or StreamField).

What are the risks of not doing so? e.g. fully replacing the `PageInterface`?

grapple/types/streamfield.py
```diff
@@ -391,20 +390,17 @@ def resolve_image(self, info, **kwargs):
         }
     )
 
-    SnippetObjectType = SnippetTypes.get_object_type()
-    if SnippetObjectType is not None:
+    class SnippetChooserBlock(graphene.ObjectType):
```
zerolab, line 393: Previously we'd only expose the snippet chooser block if we had any registered snippets.

What are the pros/cons of always exposing it?

zerolab (commented):
Left a few notes and thoughts in passing

grapple/types/interfaces.py
mgax: > What are the risks of not doing so? e.g. fully replacing the `PageInterface`?

I'm not sure. But, if I'm replacing Grapple's default page/snippet interface, I'm probably inheriting from the default one, because it feels like the right thing to do.

tests/testapp/migrations/0003_alter_blogpage_body.py
```diff
@@ -0,0 +1,126 @@
+# Generated by Django 5.0.9 on 2024-09-20 07:17
```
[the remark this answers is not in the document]
mgax: Ah, I can do that in this PR.

grapple/types/interfaces.py
```diff
@@ -209,3 +209,18 @@ def resolve_raw_value(self, info, **kwargs):
...
+
+class SnippetInterface(graphene.Interface):
+    snippet_type = graphene.String(required=True)
+
+    @classmethod
+    def resolve_type(cls, instance, info, **kwargs):
+        return registry.snippets[type(instance)]
+
+    def resolve_snippet_type(self, info, **kwargs):
+        return self.__class__.__name__
```
jams2, line 226: Should we be using the `appname.ModelName` form? Does Wagtail require snippet names to be unique?

grapple/types/streamfield.py
mgax: It looks like it was not possible to construct `SnippetObjectType` if there were no snippets defined: https://github.com/torchbox/wagtail-grapple/blob/6cebc7be1fcca437fcadd454ced3c11758da77f8/grapple/types/snippets.py#L6-L10

I don't see any downsides to always exposing the snippet chooser block. I've tried this in a minimal project and StreamField queries still work.

tests/testapp/migrations/0003_alter_blogpage_body.py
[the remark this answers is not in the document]
mgax: I've folded migration `0003` into `0001`, though I'm now wondering why it was generated, and why the `wagtail.fields.StreamField` constructor arguments have changed.

grapple/types/interfaces.py
mgax: Well, for the Page type, we have the `pageType` field, which returns the page class name, and `contentType`, which returns the `appname.ModelName` form. We could add the `contentType` field to `SnippetInterface` too.

grapple/types/interfaces.py
mgax: We can add a check, and print a warning, I guess? But it feels outside the scope of this ticket.

tests/testapp/migrations/0001_initial.py
```diff
@@ -92,193 +92,123 @@ class Migration(migrations.Migration):
                     "body",
                     wagtail.fields.StreamField(
                         [
-                            (
-                                "heading",
-                                wagtail.blocks.CharBlock(form_classname="full title"),
+                            ("heading", 0),
```
zerolab, line 95: this looks like the Wagtail 6.2 changes - https://docs.wagtail.org/en/latest/releases/6.2.html#compact-streamfield-representation-for-migrations

What version did you run the migrations with? It doesn't matter as such because they are more logical representations, and we don't do anything with SF migration

zerolab (commented):
if @jams2 is happy with it, I am happy

let's follow up with the checks on the custom interfaces as inheriting from ours

tests/testapp/migrations/0001_initial.py
mgax: Yep, I've generated them with Wagtail 6.2. But the test matrix is passing, so it looks like older Wagtail versions can still read them.

grapple/types/interfaces.py
mgax: I've created a ticket to address this: https://github.com/torchbox/wagtail-grapple/issues/408.

jams2:
@mgax looks good, thanks.

zerolab (approved):

PR #407, opened 2024-09-20
Feat/search fields

dopry:
My searchFields tests don't seem to be working quite the way I'd like. I tried setting up a body field to specify alongside title, but it didn't work. At first I didn't realize why.  In my customized grapple implementation, I use my own custom page type as the root for all queries rather the wagtail.model.Page that has all my common search_fields...  I'm realizing that doesn't work with grapple atm as you can't specify an alternate root model.   Where grapple uses a [mixin with qs=WagtailPageObjects.all()](https://github.com/torchbox/wagtail-grapple/blob/0d20a8cd8de6cceb260a1ade564effa6c2db1e79/grapple/types/pages.py#L377),  I use standalone function that looks like
```
def _resolve_search_pages(
    info,
    content_types=[],
    page=1,
    per_page=10,
    root=TnPage.objects.live().specific(),
    categories=[],
    tags=[],
    ancestor=None,
    parent=None,
    in_menu=None,
    **kwargs
):
    """
    root allows an alternative queryset to be passed in to allow
    for inheriting types to filter the queryset first.
    """
```

I need to resolve this issue before this will be usable to end users... possible solutions...

1. pass a 'root' into PagesQuery to overide the base model for pages and add a way to specify the root model.
2. find a way to get this to work with .get_specific_page

I don't have any other ideas at the moment.


dopry:
I did some additional experimenting with the testapp today. Next steps are to test query blogPages directly after specifying body as a search field on blogPages to see if that works...  If it does huzzah there is some value here that may be worth merging with proper documentation... otherwise I need to proceed with figuring out the alternate root or get specific pages issue

zerolab:
So the issue with search fields applying across different page models is that we work with the bas Page class, and even if we did specific, you can't do a simple "search across these fields" in that we need construct an OR query for each page type that has the given fields.

It should work for the specific ones like `blogPages` because that is one page type

dopry:
I believe that is the issue. I need to get around to verifying it and writing the tests for it.

PR #411, opened 2024-11-18
Support for ImageBlock

b-tupiti:
While working on a project with Next.js, GraphQL, and Wagtail, I identified that Wagtail's new ImageBlock is not currently supported, hence the need to add support for it.

### DIFF
--- a/grapple/types/streamfield.py
+++ b/grapple/types/streamfield.py
@@ -381,12 +381,22 @@ def register_streamfield_blocks():
 
         def resolve_image(self, info, **kwargs):
             return self.value
+        
+    class ImageBlock(graphene.ObjectType):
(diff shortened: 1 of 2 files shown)


grapple/types/streamfield.py
```diff
@@ -381,12 +381,22 @@ class Meta:
 
         def resolve_image(self, info, **kwargs):
             return self.value
+        
+    class ImageBlock(graphene.ObjectType):
+        image = graphene.Field(get_image_type, required=False)
```
zerolab, line 386: to check: whether we need to surface alt_text and whether this is decorative or not

zerolab:
Thanks @b-tupiti. Closing in favour of #419

b-tupiti:
Appreciate it. Thanks for the fix.

PR #415, opened 2025-02-06
Upgrade to `actions/upload-artifact@v4`; fix testing matrix

mgax:
CI runs are failing with this error:

```
Error: This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3`. Learn more: https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/
```

I've upgraded to v4 of the `upload-artifact` and `download-artifact` actions: https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md.

One change that affects this repository is about "Multiple uploads to the same named Artifact"; the uploads need to have distinct names, and then get merged during download.

### DIFF
--- a/tests/requirements.txt
+++ b/tests/requirements.txt
@@ -4,7 +4,7 @@ ruff==0.0.287
 
 # Runtime requirements
 Django>=4.2,<5.1
-wagtail>=4.1
+wagtail>=5.2,<6.3
 graphene-django>=3.0.0
 factory-boy==3.2.1
 wagtail-factories>=4.1.0
--- a/tox.ini
+++ b/tox.ini
@@ -45,13 +45,8 @@ deps =
 
     django42: Django>=4.2,<5.0
     django50: Django>=5.0,<5.1
-    wagtail52: wagtail>=5.2,<6.0
-    wagtail60: wagtail>=6.0,<6.1
-    wagtail61: wagtail>=6.1,<6.2
     wagtail62: wagtail>=6.2,<6.3
     interactive: wagtail>=5.2
-
-install_command = python -Im pip install {opts} {packages}
 
 commands =
     python -m coverage run manage.py test {posargs: -v1} --exclude-tag=needs-custom-settings
--- a/tox.ini
+++ b/tox.ini
@@ -2,8 +2,8 @@
 min_version = 4.0
 
 envlist =
-    py{38,39,310,311,312}-django{42}-wagtail{52,60}
-    py{310,311,312}-django{50}-wagtail{52,60}
+    py{38,39,310,311,312}-django{42}-wagtail{62}
+    py{310,311,312}-django{50}-wagtail{62}
     interactive
 
 [gh-actions]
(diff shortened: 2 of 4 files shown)


tests/requirements.txt
```diff
@@ -4,7 +4,7 @@ ruff==0.0.287
 
 # Runtime requirements
 Django>=4.2,<5.1
-wagtail>=4.1
+wagtail>=6.2,<6.3
```
zerolab, line 7:
```suggestion
wagtail>=5.2,<6.3
```

allthough I would even go as far as `<6.5`

tox.ini
```diff
@@ -47,9 +47,11 @@ deps =
     django50: Django>=5.0,<5.1
     wagtail52: wagtail>=5.2,<6.0
     wagtail60: wagtail>=6.0,<6.1
+    wagtail61: wagtail>=6.1,<6.2
+    wagtail62: wagtail>=6.2,<6.3
     interactive: wagtail>=5.2
 
-install_command = python -Im pip install -U {opts} {packages}
+install_command = python -Im pip install {opts} {packages}
```
mgax, line 54: The `-U` flag here was causing Wagtail to be upgraded to the latest version in [the install_package step](https://github.com/torchbox/wagtail-grapple/actions/runs/13176328127/job/36776414261?pr=415#step:6:21). You can see `wagtail==6.4` in [the output of `pip freeze`](https://github.com/torchbox/wagtail-grapple/actions/runs/13176328127/job/36776414261?pr=415#step:6:23).

tox.ini
```diff
@@ -2,8 +2,8 @@
 min_version = 4.0
 
 envlist =
-    py{38,39,310,311,312}-django{42}-wagtail{52,60}
-    py{310,311,312}-django{50}-wagtail{52,60}
+    py{38,39,310,311,312}-django{42}-wagtail{62}
+    py{310,311,312}-django{50}-wagtail{62}
```
mgax, line 6: It looks like [the last successful pipeline run](https://github.com/torchbox/wagtail-grapple/actions/runs/11017307470) on `main` was on 2024-09-24. Because of [this issue](https://github.com/torchbox/wagtail-grapple/pull/415/files#r1945018125), it was running against Wagtail 4.2. I was able to run `tox` successfully with Wagtail 4.2.

Wagtail 4.1 failed because it doesn't understand `block_lookup` in migrations: https://github.com/torchbox/wagtail-grapple/blob/86f03575fc9afe7ae317d1b4045568b5f4e78b72/tests/testapp/migrations/0001_initial.py#L114

Wagtail 4.3 failed with the error `django.db.utils.OperationalError: table testapp_customimage has no column named description`.

---
I suggest we get the CI pipeline back into a functional state, and fix Wagtail version compatibility in a subsequent PR.

tox.ini
```diff
@@ -47,9 +47,11 @@ deps =
     django50: Django>=5.0,<5.1
     wagtail52: wagtail>=5.2,<6.0
     wagtail60: wagtail>=6.0,<6.1
+    wagtail61: wagtail>=6.1,<6.2
```
zerolab, line 50:
```suggestion
```

tox.ini
zerolab: you can remove this since the default is `python -I -m pip install {opts} {packages}`

tox.ini
zerolab: well, and 5.2 since it is not in the matrix

tox.ini
mgax: If you look at the CI output, it will first install the versions in the testing matrix, and then install the package and its dependencies. Because `pip install` had the `-U` flag, this caused Wagtail to be upgraded to latest stable, regardless of what the matrix installed.

tox.ini
mgax: Ah, grand! This means I can use `tox-uv` locally, to install environments faster 🚀

zerolab: adding `tox-uv` was on my list to try with grapple, so go wild ;)

zerolab (approved):
Let's get this in and add the other changes as discussed

PR #416, opened 2025-02-07
Add support for Wagtail 6.3 and 6.4

mgax:
[Wagtail 6.3](https://docs.wagtail.org/en/stable/releases/6.3.html) has made a couple of changes to image and document models:
- Images have a new `description` field.
- Documents have their `file_size` field changed to `PositiveBigIntegerField`.

This requires a change in the migrations for the test app, which makes it incompatible with older Wagtail versions. The only older version that is still supported is Wagtail 5.2 LTS, but [only until 2025-05-01](https://github.com/wagtail/wagtail/wiki/Release-schedule#future-and-current-versions).

### DIFF
--- a/tests/settings.py
+++ b/tests/settings.py
@@ -186,13 +186,14 @@ ALLOWED_HOSTS = ["*"]
 
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 
-
-with contextlib.suppress(ImportError):
-    from .local import *  # noqa: F403
-
+# Make sure the search index is updated immediately during tests
 TASKS = {
     "default": {
         "BACKEND": "django_tasks.backends.immediate.ImmediateBackend",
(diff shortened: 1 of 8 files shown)


tests/settings.py
```diff
@@ -186,6 +186,14 @@
 
 EMAIL_BACKEND = "django.core.mail.backends.console.EmailBackend"
 
+# Make sure the search index is updated immediately during tests
+TASKS = {
+    "default": {
+        "BACKEND": "django_tasks.backends.immediate.ImmediateBackend",
+        "ENQUEUE_ON_COMMIT": False,
```
mgax, line 193: This took me way too long to figure out. I need to reach for the [release notes](https://docs.wagtail.org/en/stable/releases/6.4.html#background-tasks-run-at-end-of-current-transaction) more often 🙃

zerolab (approved):

PR #418, opened 2025-02-10
Improve tooling

mgax:
- [x] Remove ruff CI, pre-commit CI already covers ruff
- [ ] Use `tox-gh-actions` + `tox-uv`: https://hugovk.dev/blog/2024/speed-up-ci-with-uv/


.github/workflows/ruff.yml
```diff
@@ -1,22 +0,0 @@
-name: Ruff
```
zerolab, line 1: please leave this as is.
it provides nice inline annotations on PRs in case linting fails

mgax: Ah, gotcha. Then I'll just update its Ruff version, and signpost the code to keep all 3 versions of Ruff in sync 😅

zerolab: `ruff check --output-format=github grapple`

mgax: On the one hand, it's intimidating to know you're looking at _everything, in real time_. On the other hand, this probably saved me a fair amount of digging. On balance, thank you!

zerolab: 'twas pure coincidence! Just happened to check what's new in GitHub

zerolab (approved):
Lovely stuff!

PR #419, opened 2025-02-10
Support for Wagtail 4.3 ImageBlock

mgax:
https://docs.wagtail.org/en/stable/releases/6.3.html#imageblock-with-alt-text-support

This PR is based on https://github.com/torchbox/wagtail-grapple/pull/411, with:
- [x] Added fields (`decorative` and `altText`)
- [x] Tests
- [x] Update of the initial migration
- [x] Changelog entry
- [x] Documentation

### DIFF
--- a/tests/testapp/factories.py
+++ b/tests/testapp/factories.py
@@ -40,6 +40,18 @@ class ImageBlockFactory(wagtail_factories.StructBlockFactory):
     class Meta:
         model = image_blocks.ImageBlock
 
+    @classmethod
+    def _construct_struct_value(cls, block_class, params):
+        if image := params["image"]:
+            decorative = params["decorative"]
+            alt_text = params["alt_text"]
+
+            # If the image is decorative, set alt_text to an empty string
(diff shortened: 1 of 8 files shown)


tests/testapp/factories.py
```diff
@@ -28,6 +29,18 @@
 
 
 # START: Block Factories
+
+
+# TODO: Contribute upstream: https://github.com/wagtail/wagtail-factories/issues/97
+class ImageBlockFactory(wagtail_factories.StructBlockFactory):
```
zerolab, line 35: this is different from the wagtail-factories one - https://github.com/wagtail/wagtail-factories/pull/98/files#diff-c75b30a3fe8b45ca9467a1840582ff9600547e3bb8892fb2a5f7a890c36714cbR252-R270 - will there be any issues?

mgax: Ah yes. I've updated the version here to match the one in https://github.com/wagtail/wagtail-factories/pull/98 and the tests still pass.

zerolab (approved):

PR #420, opened 2025-02-14
Expand uv usage in CI

zerolab:
- uses one `--with` per extra dependency as per https://docs.astral.sh/uv/guides/scripts/#running-a-script-with-dependencies ("Multiple dependencies can be requested by repeating with `--with` option.")
- switches the coverage step to uv (and drops the toml dep)
- uses uv in the publish workflow
- tidies up the test matrix (actually test with Python 3.13 in CI, reduce variations for sqlite)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,8 +18,7 @@ env:
   TOX_TESTENV_PASSENV: FORCE_COLOR
   PIP_DISABLE_PIP_VERSION_CHECK: "1"
   PIP_NO_PYTHON_VERSION_WARNING: "1"
-  # Keep in sync with .pre-commit-config.yaml/default_language_version/python.
-  PYTHON_LATEST: "3.11"
+  PYTHON_LATEST: "3.13"
 
 jobs:
   tests-sqlite:
@@ -27,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.12']
         db: ['sqlite']
 
     steps:
@@ -60,7 +59,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
         db: ['postgres']
 
     services:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [PYTHON_LATEST]
+        python-version: ["3.13"]
         db: ['sqlite']
 
     steps:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -18,22 +18,23 @@ env:
...
+  PYTHON_LATEST: "3.13"
 
 jobs:
   tests-sqlite:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.12']
```
mgax, line 29: **Question:** how come we only need to test SQLite only with Python 3.12? Why not 3.13?

tox.ini
```diff
@@ -3,7 +3,7 @@ min_version = 4.0
 
 envlist =
     py{39,310,311,312}-django{42}-wagtail{63,64}
-    py{310,311,312}-django{50}-wagtail{63,64}
+    py{312}-django{50}-wagtail{63}
```
mgax, line 6: **Question:** Django 5.0 [supports Python 3.10 through 3.12](https://docs.djangoproject.com/en/5.1/faq/install/#faq-python-version-support), why not test all of them? And why drop `*-django50-wagtail64`?

pyproject.toml
```diff
@@ -39,7 +39,7 @@ dependencies = [
 
 [project.optional-dependencies]
 testing = [
-    "coverage[toml]>=7.2.7,<8.0",
+    "coverage>=7.2.7,<8.0",
```
mgax, line 42: **Question:** I'm curious, why was the `toml` extra being installed in the first place?

mgax (commented):
Thanks for adding `uvx` where I missed it. I have a couple of questions about the testing matrix, otherwise this looks good!

.github/workflows/ci.yml
zerolab: brainfart. I reduced the sqlite matrix before I added 3.13. It should be 3.13

tox.ini
zerolab: saving some compute cycles. 5.0 reaches EOL in 2.5 months - https://endoflife.date/django - so may as well keep it minimal

pyproject.toml
zerolab: at one point the coverage config was in pyproject.toml. That changed in #388 where I forgot to keep it simple

.github/workflows/ci.yml
zerolab: 8fd932a

pyproject.toml
mgax: Got it, thanks!

tox.ini
mgax: Ah, interesting, makes sense 👍

mgax (approved):

.github/workflows/ci.yml
```diff
@@ -18,22 +18,23 @@ env:
...
+  PYTHON_LATEST: "3.13"
 
 jobs:
   tests-sqlite:
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: [PYTHON_LATEST]
```
mgax, line 29: It looks like this is not getting interpolated: https://github.com/torchbox/wagtail-grapple/actions/runs/13347176830/job/37279042624?pr=420#step:3:16.

```suggestion
        python-version: [$PYTHON_LATEST]
```

zerolab: something, something.. copy/pasta fail

PR #422, opened 2025-03-21
Bugfix: Add WAGTAILADMIN_BASE_URL in the get_media_item_url function

Hercilio1:
The current Wagtail version removed `BASE_URL`, so to ensure the `get_media_item_url` function works, I updated it to try both, prioritizing `WAGTAILADMIN_BASE_URL`.

I'm having issues with my current project when trying to get the image URL using the following GraphQL query:
```
{
  pages {
    ... on PageForTests {
      id
      title
      body {
        ... on CustomImageBlock {
          id
          altText
          image {
            url
          }
        }
      }
    }
  }
}
```


jams2:
Same issue for me when running the Django dev server behind a proxy. I had to set BASE_URL for image srcset URLs to be correct.

PR #429, opened 2026-04-20
Add `isDefaultSite` filter to sites query

ianmeigh:
Adds an optional `isDefaultSite: Boolean` argument to the `sites` GraphQL query, allowing sites to be filtered by their default status.

This is useful for multi-site Wagtail setups where we want to query only the default site for things like global nav, settings, or fallback content, without knowing its `hostname` or `id`.


jams2 (changes requested):
Thanks @ianmeigh . Could you please update the changelog and the [documentation](https://github.com/torchbox/wagtail-grapple/blob/main/docs/general-usage/graphql-types.rst?plain=1#L158-L175) - we should document the query arg as shown in the `site` query example.

docs/general-usage/graphql-types.rst
```diff
@@ -155,12 +155,19 @@ the ``sites`` or ``site`` field on the root query type. Available fields for the
...
+    # Optionally filters sites by their default site status. Omitting this returns all sites.
+    is_default_site: Boolean
+
+and can be queried like so:
 
 ::
 
     query {
-        sites {
+        sites (isDefaultSite: True) {
```
jams2: Issue: should be lowercase `true`

jams2 (changes requested):

docs/general-usage/graphql-types.rst
ianmeigh: Changed in https://github.com/torchbox/wagtail-grapple/pull/429/commits/341ef51b6ad52947a1ba41147753c235cf9326d9.

jams2 (approved):

PR #430, opened 2026-05-23
Fix DjangoObjectType deprecation warnings

### LINKED ISSUE #355: Creating a DjangoObjectType without either the `fields` or the `exclude` option is deprecated
Upgrading to wagtail 5.1 and graphene 3.3 results in the following deprecation warnings for all models registered with wagtail-grapple:
>/home/wagtail/.local/lib/python3.9/site-packages/graphene/utils/subclass_with_meta.py:46: DeprecationWarning: Creating a DjangoObjectType without either the `fields` or the `exclude` option is deprecated. Add an explicit `fields = '__all__'` option on DjangoObjectType StubBlogPost to use all fields

The issue appears to be coming from [grapple.actions.build_node_type](https://github.com/Morsey187/wagtail-grapple/blob/dfb1c2211f9636a5a1c85e4ad708089812800c47/grapple/actions.py#L263-L264):
```
    class StubMeta:
        model = stub_model
```
Based on the warning, there should be a `fields = '__all__'` attribute included.

ponas:
The following warnings were emitted from `graphene-django`:

```
DeprecationWarning: Creating a DjangoObjectType without either the `fields` or the `exclude` option is deprecated. Add an explicit `fields = '__all__'` option on DjangoObjectType CollectionObjectType to use all fields
DeprecationWarning: Defining `exclude_fields` is deprecated in favour of `exclude`.
```

Fixed by:
* explicitly setting `fields = "__all__"`
* setting `exclude` instead of `exclude_fields`

Closes #355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Unreleased
 
+### Fixed
+
+- Declare explicit `fields`/`exclude` on `DjangoObjectType` subclasses to silence graphene-django deprecation warnings ([#430](https://github.com/torchbox/wagtail-grapple/pull/430)) @ponas
+
 ## [0.31.0] - 2026-04-21
 
 ### Added
--- a/grapple/actions.py
+++ b/grapple/actions.py
@@ -312,7 +312,7 @@ def load_type_fields():
 
                 type_meta = {"Meta": Meta, "id": graphene.ID(), "name": type_name}
 
-                exclude_fields = []
+                exclude = []
                 base_type_for_exclusion_checks = (
                     base_type if not issubclass(cls, WagtailPage) else WagtailPage
                 )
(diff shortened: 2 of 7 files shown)


jams2:
Hi @ponas , thanks for this.

I tested your branch in my project, it silences the warnings, and does not cause a change in the schema.

Relevant docs for the record:

* https://docs.graphene-python.org/projects/django/en/latest/queries/#specifying-which-fields-to-include
* https://docs.graphene-python.org/projects/django/en/latest/queries/#exclude

Could you please add a changelog entry?

ponas:
Thanks for dealing with the changelog entry!

As a note I also tested this branch on a fairly large project with no schema changes. Seems to be a safe change.

jams2:
Thanks @ponas !

